PR #14, opened 2023-07-25
Pass initial test suite

123epsilon:
Allows retrieving the element size of a Tensor based on its datatype, this passes the initial test suite.

### DIFF
--- a/python/shark_turbine/dynamo/tensor.py
+++ b/python/shark_turbine/dynamo/tensor.py
@@ -40,6 +40,23 @@ from iree.runtime import (
 # Factories and device enablement
 ###############################################################################
 
+def get_itemsize(dtype: torch.dtype):
+    """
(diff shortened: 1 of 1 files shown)


python/shark_turbine/dynamo/tensor.py
```diff
@@ -40,6 +40,23 @@
 # Factories and device enablement
 ###############################################################################
 
+def get_itemsize(dtype: torch.dtype):
```
stellaraccident, line 43: Fine to do in a followup: There are some dtype tables at the bottom. Can you move this down there and make it into a dict lookup?

stellaraccident (approved):

python/shark_turbine/dynamo/tensor.py
123epsilon: Ah I missed that detail - I'll move it over.

PR #16, opened 2023-07-28
Parse list and scalar args for torch ops

123epsilon:
Resolves an issue with the `FxImporter` where arguments that are `!torch.list` types are not properly imported due to the Fx tracer denoting them via `immutable_list` types rather than encapsulating them in an `fx.Node`. Also addresses a case where a python scalar is not parsed into an `fx.Node` but rather retains its python type. For instance, in [`aten.convolution`](https://github.com/llvm/torch-mlir/blob/0109bf705b0dae3b4d55c7a9b180516787cbeac6/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td#L4583C18-L4583C18) where the "transposed" parameter is passed as false by default - this argument is left as a `bool` and therefore needs to be handled when importing arguments.

### DIFF
--- a/python/shark_turbine/dynamo/importer.py
+++ b/python/shark_turbine/dynamo/importer.py
@@ -335,7 +335,7 @@ class GraphNodeImporter:
                 elif op == "output":
                     # args[0] is a singleton tuple that we flatten into multiple
                     # results.
-                    operands = [self._import_argument(arg) for arg in node.args[0]]
+                    operands = [self._import_argument(loc, arg) for arg in node.args[0]]
                     func_dialect.ReturnOp(operands, loc=loc)
 
     def _import_torch_op_overload(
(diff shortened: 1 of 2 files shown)


python/shark_turbine/dynamo/importer.py
```diff
@@ -401,14 +401,37 @@ def _import_torch_op_overload(
         for i, value in enumerate(operation.results):
             self._v[(node, i)] = value
 
-    def _import_argument(self, arg: NodeArgument) -> Value:
+    def _import_argument(self, arg: NodeArgument, loc: Location = None) -> Value:
```
stellaraccident, line 404: Nit: In MLIR by convention, we typically always include the Location first in the argument list for cases like this and make it non-optional.

stellaraccident (approved):
One small nit.

python/shark_turbine/dynamo/importer.py
123epsilon: Sounds good, I've pushed an edit

stellaraccident (approved):

PR #17, opened 2023-08-01
Fix importing torch.device as arg

123epsilon:
Addresses an import issue when either implicitly or explicitly giving `device` as an argument to a torch op, here we simply create a mapping to construct a `torch.constant.device` op when we encounter a `torch.device` argument.


python/test/dynamo/importer_basic_test.py
```diff
@@ -44,6 +44,27 @@ def basic(x):
 
         basic(torch.randn(3, 4))
 
+    def testImportDevice(self):
```
stellaraccident, line 47: You're accumulating a handful of these tests that do the same basic thing. When you have time, would be good to modularize them a bit so the test itself is using common code for import and mostly just has what is being tested.

stellaraccident (approved):

python/test/dynamo/importer_basic_test.py
123epsilon: Yeah, we'll include that in the next PR to clean up the tests a bit

brucekimrokcmu: should be reflected in the next PR :)

PR #18, opened 2023-08-01
Fix importing torch dtypes

123epsilon:
This PR implements proper parsing of `torch.dtype` subclasses to the appropriate `Torch_IntValue` using torch-mlir convention, taken from [here](https://github.com/llvm/torch-mlir/blob/4c24472dea1c9102b898768b0b11e31487e50207/python/torch_mlir/_dynamo_fx_importer.py#L189). We do not support quantized datatypes due to dynamo's lack of support for quantized datatypes outside of a set of specific backends, but may address these cases in the future.


stellaraccident (approved):
Thanks!

PR #20, opened 2023-08-02
Fix importing multiple returns and intervening to choose right Scalar ops

brucekimrokcmu:
This PR resolves the following two issues and modularizes importer call via `create_backend()`.
- Implementation to unpack multiple aten results return when default `getitem` doesn't work due to `node.args` data structure 
- Inadequate ops return from AOT Autograd, when the second argument is scalar for the following ops.
`"torch.aten.mul.Tensor", "torch.aten.div.Tensor", "torch.aten.add.Tensor", "torch.aten.sub.Tensor", "torch.aten.floor_divide"`

For the first issue, we simply parse meta data 'val' and store them as tuple to be passed as `result_types` for `Operation.create(...)`.

For the second issue, we intervene to chose the scalar version of the ops, when the ops have scalar as 2nd argument. Note that in case of `floor_div` operation, failure in legalizing `torch.constant.float` or `torch.constant.int` ops during lowering occur due to partial implementation of floor_div op in torch-mlir. Therefore, we added a `tensor_scalar_op_conversion_importer_test.py` to show that the importer works with a comment that more work is needed from torch-mlir side.


python/shark_turbine/dynamo/importer.py
```diff
@@ -373,6 +373,18 @@ def _import_torch_op_overload(
         if schema.overload_name != "":
             mlir_op_name += f".{schema.overload_name}"
 
+        # Aot_autograd is not returning scalar versions of the below ops, when the second argument is scalar.
+        # Thus, intervening to select the right ops.
+        if mlir_op_name in ["torch.aten.mul.Tensor", "torch.aten.div.Tensor", "torch.aten.add.Tensor",
```
stellaraccident, line 378: Can you bust this list out as a module level constant and an appropriate comment?

(when you do that, might as well store it as a `set()` for efficiency of lookup)

stellaraccident: Actually, I would probably make this a dict so that I could map the original op name to the revised op name and avoid the special case rewriting below (i.e. just do a `get()` on the dict and use the substition if not None).

stellaraccident (approved):
Nice. A couple of minor comments to address before landing.

Can you also run `black python/` before landing to make sure formatted?

123epsilon (approved):

python/shark_turbine/dynamo/importer.py
```diff
@@ -373,6 +373,13 @@ def _import_torch_op_overload(
         if schema.overload_name != "":
             mlir_op_name += f".{schema.overload_name}"
 
+        # Intervening to use Scalar ops due to incorrect ops from AOT-autograd with scalar arguments.
+        if mlir_op_name in TENSOR_SCALAR_OP_CONVERTER:
+            if isinstance(node.args[1], float) or isinstance(node.args[1], int):
+                mlir_op_name = TENSOR_SCALAR_OP_CONVERTER.get(
```
stellaraccident, line 379: Nit: Sightly more efficient to use [] instead of get if you know it won't raise an exception.

python/shark_turbine/dynamo/importer.py
```diff
@@ -373,6 +373,13 @@ def _import_torch_op_overload(
         if schema.overload_name != "":
             mlir_op_name += f".{schema.overload_name}"
 
+        # Intervening to use Scalar ops due to incorrect ops from AOT-autograd with scalar arguments.
+        if mlir_op_name in TENSOR_SCALAR_OP_CONVERTER:
```
stellaraccident, line 377: Nit: for performance sensitive code, I usually go to the trouble of not doing two lookups: out it in a temporary with .get and then condition in not none.

python/shark_turbine/dynamo/importer.py
```diff
@@ -573,5 +593,13 @@ def _make_constant_op(
 SCALAR_TYPE_TO_TORCH_TYPE = {
     int: "!torch.int",
     float: "!torch.float",
-    str: "!torch.str"
-}
\ No newline at end of file
+    str: "!torch.str",
+}
+
+TENSOR_SCALAR_OP_CONVERTER = {
```
stellaraccident, line 599: A comment please on what this is and that we may want to remove it some day.

python/shark_turbine/dynamo/importer.py
brucekimrokcmu: Thanks for the feedback! Super helpful for my growth :)

bailuan:
Hi, have you checked the model including "unbind + getitem" ?
I found that maybe turbine doesn't have a good support for getitem when it follows "unbind" node.
In my test case, the output's data struct of aten.unbind is tuple(tensor), but not list(tensor). It looks like a multiple  outputs, but actually it is fake, torch think it is a single output. So the "unbind" node would not been added to "_multi_result_nodes", which raising "NotImplementedError( f"General getitem access to non-multi-result ops")".
I am not very familiar with shark-turbine, can you tell me what should I do?

PR #21, opened 2023-08-03
Import torch enum args

123epsilon:
This PR addresses importing torch enum types (`torch.dtype`, `torch.layout`, `torch.memory_format`) by converting them to the proper integer values.

### DIFF
--- a/python/shark_turbine/dynamo/importer.py
+++ b/python/shark_turbine/dynamo/importer.py
@@ -123,8 +123,6 @@ TORCH_LAYOUT_TO_INT = {
     torch.sparse_bsc: 5,
 }
 
-TORCH_ENUM_TYPES = {torch.dtype, torch.layout, torch.memory_format}
-
 
 class FxImporter:
     """Main entry-point for importing an fx.GraphModule."""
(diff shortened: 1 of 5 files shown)


python/shark_turbine/dynamo/importer.py
```diff
@@ -497,6 +509,35 @@ def _import_default_value(self, loc: Location, arg, expected_jit_type) -> Value:
...
 
+    def _import_enum_value(self, loc: Location, arg):
+        """Import the correct integer representation of torch enum values"""
+        assert (
+            type(arg) in TORCH_ENUM_TYPES
+        ), f"Expected enum type to be one of {TORCH_ENUM_TYPES}, got {type(arg)}"
+        mapping = None
+        enum_cls = None
+
+        if isinstance(arg, torch.dtype):
```
stellaraccident, line 520: Nit: the TypeSubclassMap below is intended to be a more efficient way to manage such cascaded isinstance conditions (by turning them from a cascade of expensive lookups to a direct hash lookup that can be specified in a module level global vs in code.

Have a look and fine to adapt in a followup.

stellaraccident (approved):

python/shark_turbine/dynamo/importer.py
123epsilon: Good point, I thought to keep them logically separated but that makes sense. I've pushed a commit refactoring it to use the TypeSubclassMap

PR #23, opened 2023-08-04
Accept default list arguments

123epsilon:
This refactors the list construction logic into a new importer method `_import_list_argument`, cases where users explicitly pass lists were handled before in `_import_argument` but when the list arguments are instead filled by default value we have to use essentially the same logic to construct the list. 

For example:
```py
def foo():
    m = nn.MaxPool2d(2, 2, 1) # params like 'dilation' are defaulted to a constant list: [1,1]
    return m(torch.randn(1,256,256))
```

I implemented this as a class method because the logic for most other cases requires access to the importer's state, and creating a separate method to call from `LITERAL_CONVERTER_MAP` to handle the case with lists made purely of constant values would repeat code (and require breaking the signature convention we are using for this lookup table), but I can edit this as needed. 

I also didn't include a test case here because this really reuses existing logic, but if we think it's needed I can include one.

### DIFF
--- a/python/shark_turbine/dynamo/importer.py
+++ b/python/shark_turbine/dynamo/importer.py
@@ -461,44 +461,7 @@ class GraphNodeImporter:
                 raise RuntimeError(f"Attempt to de-reference a multi-result node")
             return self._v[(arg, 0)]
         elif isinstance(arg, torch_fx.immutable_collections.immutable_list):
-            # create list operands
-            list_operands = []
-            result_type = MlirType.parse(
-                SCALAR_TYPE_TO_TORCH_LIST_TYPE[type(arg[0])], context=self._c
-            ) if type(arg[0]) in SCALAR_TYPE_TO_TORCH_LIST_TYPE else None
-
-            for operand in arg:
-                if not isinstance(operand, type(arg[0])):
-                    raise TypeError(
-                        f"Lists with multiple types are not supported, got: {type(arg[0])}, {type(operand)}"
-                    )
-
-                if isinstance(operand, torch.fx.Node):
-                    if operand in self._multi_result_nodes:
-                        raise RuntimeError(f"Attempt to de-reference a multi-result node")
-                    val = self._v[(operand,0)]
-                    if result_type is None:
-                        list_type: str = str(val.type)
-                        begin_index = 7 if list_type.startswith("!torch.") else None
-                        end_index = list_type.find("<")
-                        end_index = end_index if end_index != -1 else None
-                        list_type = list_type[begin_index:end_index]
-                        result_type = MlirType.parse(f"!torch.list<{list_type}>")
-                else:
-                    val = self._import_default_value(
-                        loc, operand, SCALAR_TYPE_TO_TORCH_TYPE[type(operand)]
-                    )
-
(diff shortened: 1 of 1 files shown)


python/shark_turbine/dynamo/importer.py
```diff
@@ -461,53 +461,59 @@ def _import_argument(self, loc: Location, arg: NodeArgument) -> Value:
...
             with loc:
                 arg_value = LITERAL_CONVERTER_MAP.lookup(type(arg))(arg, self, self._cc)
             return arg_value
         else:
             raise NotImplementedError(f"FIXME: Unsupported Node Argument: {arg}")
 
+    def _import_list_argument(self, loc: Location, arg):
+        # create list operands
+        list_operands = []
+        result_type = MlirType.parse(
```
stellaraccident, line 475: Can you refactor this to an if/else? These multi-line ternary expressions are really hard to read at a glance.

stellaraccident: Also, please eliminate redundancy of lookups. `type(arg[0])` is not free and neither is the dict lookup.

python/shark_turbine/dynamo/importer.py
```diff
@@ -461,53 +461,59 @@ def _import_argument(self, loc: Location, arg: NodeArgument) -> Value:
...
+                raise TypeError(
+                    f"Lists with multiple types are not supported, got: {type(arg[0])}, {type(operand)}"
+                )
+
+            if isinstance(operand, torch.fx.Node):
+                if operand in self._multi_result_nodes:
+                    raise RuntimeError(f"Attempt to de-reference a multi-result node")
+                val = self._v[(operand,0)]
+                if result_type is None:
+                    list_type: str = str(val.type)
```
stellaraccident, line 490: I don't love that we are text parsing the type here but I'll not resist too much for now. However, can you please at least use a regex vs this indexy string manipulation?

python/shark_turbine/dynamo/importer.py
123epsilon: I'll refactor this

python/shark_turbine/dynamo/importer.py
123epsilon: Yes, I can do that. Ideally, I'd like to use something like `torchMlirTorchListTypeGet` from [here](https://github.com/llvm/torch-mlir/blob/9a1fae97b1e56464a109206acd3b4be7d63ebfc7/lib/CAPI/TorchTypes.cpp#L132) but I'm not sure how I can access that from the python bindings available in turbine at the moment - do you have any idea how I can go about this? I don't see any bindings generated for torch-mlir (i.e. `_torch_ops_gen.py`) in `iree.compile` by which I might access the torch-mlir `ListType`.

123epsilon: String indexing is now refactored as regex

stellaraccident: I was hoping we wouldn't need most of those because they are complicated and slow with all of the goo (which will make it harder to build a faster interface later).

This importer isn't productivity focused, so I'm fine if it has some hacks that we can clean up later.

PR #25, opened 2023-08-09
Test simple backward

brucekimrokcmu:
We implemented the followings for this PR:
- Added `importer_backward_test.py', which currently has a simple custom loss function for backward. 
Needs to add test cases including 1) using `func.grad` for backward test, and 2) MNIST Classifier using LeNet for backward test
- Parsing `!torch.none` to store MlirType of none for `result_types` to create required op
- Added `bool` for list construction. 

Test cases mentioned above will be added in the next PR.

### DIFF
--- a/python/shark_turbine/dynamo/importer.py
+++ b/python/shark_turbine/dynamo/importer.py
@@ -351,8 +351,6 @@ class GraphNodeImporter:
                     self._v[(node, 0)] = self._b.arguments[num_placeholders]
                     num_placeholders += 1
                 elif op == "call_function":
-                    # print("NODE", node
-                    # print("NODEMETA", node.meta)
 
                     target = node.target
                     if target == operator.getitem:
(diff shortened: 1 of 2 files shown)


python/shark_turbine/dynamo/importer.py
```diff
@@ -152,6 +153,7 @@ def __init__(
             self._config_check()
         self._cc = ContextCache(self._c)
         self._m_ip = InsertionPoint(self._m.body)
+        self.used_function_names = set()
```
stellaraccident, line 156: Can you include a pytype annotation: `self.used_function_names: Set[str] = set()`

python/shark_turbine/dynamo/importer.py
```diff
@@ -171,6 +173,11 @@ def import_graph_module(self, gm: GraphModule):
         self.import_stateless_graph(gm.graph)
 
     def import_stateless_graph(self, g: Graph, func_name: str = "main"):
+        if func_name in self.used_function_names:
```
stellaraccident, line 176: Can you add: `TODO(stella): Switch this to SymbolTable insertion/dedup`

stellaraccident (approved):
Couple of small nits then lgtm.

PR #27, opened 2023-08-10
Add decompositions to import flow

123epsilon:
Adds default decompositions from SHARK to the Turbine import flow, with a test for importing `aten.chunk` which previously failed.

A few design questions:
1) Here I've included all the default decompositions present in SHARK, do we want to keep it this way or should I by default include a more minimal set?

2) I've hardcoded the above set of default decompositions for our CPU backend, but we spoke about making decompositions configurable - do we want a nice, user-facing way to do this or only want this for development purposes? I'm trying to figure out the best way to include this feature without it seeming awkward

Also, adding default decompositions to the generated tests passes an extra 50 cases when considering the first 700 or so test cases (this equates to setting `--limit 100` when running the test suite)

### DIFF
--- a/python/shark_turbine/dynamo/backends/cpu.py
+++ b/python/shark_turbine/dynamo/backends/cpu.py
@@ -37,6 +37,9 @@ from ..importer import FxImporter
 
 import torch
 from torch._dynamo.backends.common import aot_autograd
+from torch.fx.experimental.proxy_tensor import make_fx
+from torch._decomp import get_decompositions
+from torch.func import functionalize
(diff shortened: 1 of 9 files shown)


123epsilon:
Edit: added another decomposition that handles batchnorm2d as it is very widespread and resolved an additional 200-300 test cases in our test suite (again limiting ourselves to the first 100 modules which is ~700 testcases).

stellaraccident (approved):
Good enough for now. Thanks.

PR #28, opened 2023-08-11
Implemented symbol table to prevent redefinition of func symbol name

brucekimrokcmu:
Implemented symbol table to prevent redefinition of symbol named during backward(). `loss.backward()` in an example test case will call `import_stateless_graph()` again, and thus requires func symbol name inside module not to be redefined.
Now with the symbol table implemented, we no longer check against a set() that stores previously used func names as we used to do in PR #25.

After `FuncOp` is initially created in the FX Importer, `self.symbol_table.insert(func)` insertion renames this symbol in the symbol table.

### DIFF
--- a/python/shark_turbine/dynamo/importer.py
+++ b/python/shark_turbine/dynamo/importer.py
@@ -3,7 +3,7 @@
 # Licensed under the Apache License v2.0 with LLVM Exceptions.
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-
+import builtins
 import logging
 import operator
 import re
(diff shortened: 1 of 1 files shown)


python/shark_turbine/dynamo/importer.py
```diff
@@ -191,6 +185,7 @@ def import_stateless_graph(self, g: Graph, func_name: str = "main"):
             entry_block = Block.create_at_start(func.body, ftype.inputs)
         node_importer = GraphNodeImporter(self._c, self._cc, entry_block)
         node_importer.import_nodes(g.nodes)
+        self.symbol_table.insert(func)
```
stellaraccident, line 188: Note that inserting into the symbol table can change the function's symbol, so you need to make sure that anywhere you were using the original name doesn't escape.

stellaraccident: (this looks fine to me, just noting)

stellaraccident (approved):

PR #29, opened 2023-08-11
Document known issues

123epsilon:
This includes a markdown file with documentation for some known issues. The three included here comprise the vast majority of all errors that we encounter when attempting to import novel `nn.Module` instances. Each one also includes a minimal reproducible example. The vast majority of other errors (at least in the tests that I have sampled) have to do with unimplemented ops which can either be dealt with via including decompositions or implementing ops in upstream torch-mlir. 

Note: I included a small tweak to the importer, adding the ability to convert `None` to the appropriate `!torch.none` in our `TypeSubclassMap` because it 1) obscures the real issue in one of these cases and 2) is probably something we want there anyways.

### DIFF
--- a/python/shark_turbine/known_issues.md
+++ b/python/shark_turbine/known_issues.md
@@ -1,0 +1,54 @@
+# Known Issues in SHARK-Turbine
+
+## Handling lists of optional types
+```py
+from torch import nn
+class foodmod(nn.Module):
+    def __init__(self):
+        self.up = nn.Upsample(scale_factor=2, mode='bilinear', align_corners=True)
+    def forward(self, x):
+        return self.up(x)
+```
+```
+# occuring in importer -> import_list_arguments
+compiler_fn raised TypeError: Lists with multiple types are not supported, got: <class 'NoneType'>, <class 'torch.fx.node.Node'>
+```
+An example is attempting to import `nn.Upsample`. This module internally makes a call to `F.interpolate` which eventually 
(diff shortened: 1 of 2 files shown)


python/shark_turbine/known_issues.md
```diff
@@ -0,0 +1,54 @@
...
+from torch import nn
+class foodmod(nn.Module):
+    def __init__(self):
+        self.up = nn.Upsample(scale_factor=2, mode='bilinear', align_corners=True)
+    def forward(self, x):
+        return self.up(x)
+```
+```
+# occuring in importer -> import_list_arguments
+compiler_fn raised TypeError: Lists with multiple types are not supported, got: <class 'NoneType'>, <class 'torch.fx.node.Node'>
```
stellaraccident, line 14: Note: we usually call the heterogenous lists.

python/shark_turbine/known_issues.md
```diff
@@ -0,0 +1,54 @@
...
+# occuring in importer -> import_list_arguments
+compiler_fn raised IndexError: list index out of range
+```
+
+Currently, we have issues dealing with functional variants of
+torch operations that do not define meaningful defaults for their arguments.
+Two common operations for which this issue arises are `F.avg_pool2d` and `F.max_pool2d`.
+Taking `max_pool2d` as an example, the [functional version](https://pytorch.org/docs/stable/generated/torch.nn.functional.max_pool2d.html) sets `stride=None` by default (which returns an empty list to the importer), 
+however, the actual intended default setting is to set `stride=kernel_size`. This issue does not occur with the corresponding `nn.Module` wrapper `MaxPool2d` because
+it actually [manually sets the intended default value](https://pytorch.org/docs/stable/_modules/torch/nn/modules/pooling.html#_MaxPoolNd). The same issue is at play in `avg_pool2d`.
```
stellaraccident, line 39: Ugh.

stellaraccident (approved):

PR #36, opened 2023-08-22
Add llama test

123epsilon:
Add a test case for passing llama through the`turbine_cpu` backend. This replaces all fairscale layers with corresponding vanilla torch layers for simplicity, but we can add these back later once we have llama working. Also removes the `@torch.inference_mode()` decorator to avoid the issue documented [here](https://github.com/pytorch/pytorch/issues/101151), which is not necessarily relevant to the quality of our pipeline.


stellaraccident (approved):
Thank you.

PR #37, opened 2023-08-24
Address lift fresh copy

123epsilon:
This PR implements a workaround to the issues we face as a result of lift_fresh_copy. This op is generated to guarantee the freshness invariant of a new tensor (see [here](https://github.com/pytorch/pytorch/blob/8c62f01cb7b75cb647882a9bf85f10031b08896e/torch/fx/experimental/proxy_tensor.py#L329)). FX actually takes constant tensor objects from operations like `torch.Tensor([2.])` and [creates uniquely named attributes](https://github.com/pytorch/pytorch/blob/36399d067a56d9875fd9c2bf61434126398ffb87/torch/fx/graph.py#L843C56-L843C56) in the top level module to store their values, hence these values never touch Turbine's importer (this is also why there is an apparent graph break at these operations, they're referencing something that wasn't created in the user-defined function). Dynamo then represents these constant values with a `get_attr` node that seeks to retrieve the object associated with that name from the top level module.

The fix implemented here is simple: catch when this occurs, grab the actual object from the top level module, and create an entry in our importer's `_v` table so that we can properly associate the identifier generated by FX with the appropriate `MlirValue`. To do this we construct a `torch.vtensor.literal` op which requires an `ElementsAttr` as input. `ElementsAttr` requires a python buffer to create the corresponding mlir `tensor` object. However, there is an issue: torch `Tensor` objects do not currently implement the full python array interface so they can't be directly converted into a python buffer (see this [issue](https://github.com/pytorch/pytorch/issues/51156) and [this article](https://pearu.github.io/array_interface_pytorch.html)). So we have to convert the tensor to numpy in order to get a buffer version of our tensor. 

It's possible I overcomplicated this last part, and it has the unfortunate side effect that we can't do this for `bfloat16` (because this type doesn't exist in numpy) or `torch.complex` datatypes (because the numpy buffer format for these objects is incompatible with the buffer format `DenseElementsAttr` is expecting for complex datatypes - so if there's a better way to do this I'd love to implement it that way instead.

Generated tests are now at 90% passing.

### DIFF
--- a/python/shark_turbine/dynamo/importer.py
+++ b/python/shark_turbine/dynamo/importer.py
@@ -9,6 +9,7 @@ import operator
 import re
 from types import NoneType
 from typing import Any, Dict, List, Optional, Sequence, Set, Tuple
+import numpy as np
 
 from iree.compiler.ir import (
     Attribute as MlirAttribute,
@@ -25,6 +26,7 @@ from iree.compiler.ir import (
     StringAttr,
     Type as MlirType,
     Value,
+    DenseElementsAttr,
 )
 
 import iree.compiler.dialects.func as func_dialect
@@ -85,6 +87,24 @@ TORCH_DTYPE_TO_MLIR_TYPE_ASM = {
     torch.complex32: "complex<f16>",
     torch.complex64: "complex<f32>",
     torch.complex128: "complex<f64>",
+}
+
+TORCH_DTYPE_TO_NPY_TYPE = {
+    # torch.qint8: None, # no equivalent np datatype
+    # torch.quint8: None,
+    torch.uint8: np.uint8,
+    torch.int8: np.int8,
+    torch.int16: np.int16,
+    torch.int32: np.int32,
+    torch.int64: np.int64,
+    # torch.bf16: None, there's no equivalent np datatype so this isn't supported right now
+    torch.float16: np.float16,
+    torch.float32: np.float32,
+    torch.float64: np.float64,
+    torch.bool: np.bool_,
+    # torch.complex32: None, # no equivalent precision for numpy
+    # torch.complex64: np.complex64, # complex dtypes can't be parsed by DenseElementsAttr in the numpy buffer format
+    # torch.complex128: np.complex128,
 }
 
 # https://github.com/llvm/torch-mlir/blob/4c24472dea1c9102b898768b0b11e31487e50207/python/torch_mlir/_dynamo_fx_importer.py#L189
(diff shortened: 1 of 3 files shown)


python/shark_turbine/known_issues.md
```diff
@@ -21,60 +21,21 @@ it actually [manually sets the intended default value](https://pytorch.org/docs/
...
-```
-AssertionError: f{func} should not have fake inputs without constants
-```
+the graph. Torch dynamo dynamically creates this attribute in the top level module object, hence this object is never 
+passed through our importer, meaning that our lookup for the appropriate MlirValue in the importer's `_v` table fails. This consistently
+occurs when the graph generates an intermediate `aten.lift_fresh_copy` as in the case of creating a new tensor above.
 
+We now have a fix for this by directly instantiating the object using a reference to the top level graph module in the 
+importer, but this method does not support all torch datatypes - in particular it fails to support `bfloat16` and 
+complex datatypes.
```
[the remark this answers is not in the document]
123epsilon: Added this back with some links to relevant snippets

python/shark_turbine/dynamo/importer.py
```diff
@@ -617,6 +647,26 @@ def _make_constant_op(
...
+def _make_vtensor_literal_op(tensor: torch.Tensor, mlir_type: MlirType) -> Operation:
+    npy_dtype = TORCH_DTYPE_TO_NPY_TYPE.get(tensor.dtype)
+    assert npy_dtype is not None, f"Can not create literal tensor for unsupported datatype: {tensor.dtype}"
+    # We need a raw buffer of data in order to create an ElementsAttr for the invocation of torch.vtensor.literal,
+    # but torch.Tensor does not fulfill the python buffer/array interface hence we must convert to a numpy array to get
+    # a raw buffer of our data. We can't call torch.Tensor.numpy() directly because this internally forces a call to
+    # detach() which throws an error as we are operating in a FakeTensorMode, hence the simplest way to get this raw
+    # buffer is via the indirection: Tensor -> list -> numpy array. This allows us to create a vtensor literal as
+    # desired, but also limits which data types we can support in this function (see TORCH_DTYPE_TO_NPY_TYPE above)
+    np_tensor = np.array(tensor.tolist()).astype(npy_dtype)
```
stellaraccident, line 659: Oof. This is going to come back to bite us. Would you mind capturing the context here in an issue and noting it in the PR description to tie together? This kind of stuff is always like walking on broken glass a bit and we'll need to revisit at some point.

stellaraccident (approved):
Oh man - that is annoying. Thank you for nailing it down.

python/shark_turbine/dynamo/importer.py
123epsilon: Sounds good I'll write it up tomorrow morning, I wish there was a better way to get the buffer representation of the tensor but so far there is not -  if someone were to finish implementing the python array interface for tensors in pytorch it would give us a first-class mechanism for getting the buffer representation and we could avoid this.

This is an issue tracking this feature: https://github.com/pytorch/pytorch/issues/54138

Alternatively, we could manually do some sort of restructuring from numpy format to the elementsattr buffer format, not sure where to find those docs at the moment though, could be worth finding them.

PR #46, opened 2023-09-12
Replace lift_fresh_copy with aten.clone

123epsilon:
Previously we had a fix that allowed us to pass `lift_fresh_copy` through the importer. However, this fails to actually lower to `linalg` and then `arith` due to an issue in `torch-mlir` where `lift_fresh_copy` causes a failure in the corresponding lowering passes when using torch dynamo, this is a quick fix that replaces this op with an equivalent one: `aten.clone`. This is valid because this is actually the canonical decomposition of `lift_fresh_copy` in `torch-mlir` anyway.


stellaraccident (approved):
That op just keeps giving... This doesn't look dissimilar to what Inductor does as a WAR.

PR #47, opened 2023-09-13
Promote scalar arguments to tensors when appropriate

123epsilon:
Handles a case where there is no `Scalar` equivalent for a torch op which accepts scalar arguments - instead Pytorch promotes such scalars to a 0-rank tensor, which we now do as well.

### DIFF
--- a/python/shark_turbine/dynamo/importer.py
+++ b/python/shark_turbine/dynamo/importer.py
@@ -581,6 +581,12 @@ class GraphNodeImporter:
             isinstance(node.args[1], float) or isinstance(node.args[1], int)
         ):
             mlir_op_name = TENSOR_SCALAR_OP_CONVERTER[mlir_op_name]
+            # replace schema to retrieve the right function signature
+            op_attrs = mlir_op_name.split(".")
+            op_overload = getattr(torch, "ops")
+            for i in range(1, len(op_attrs)):
+                op_overload = getattr(op_overload, op_attrs[i])
+            schema = op_overload._schema
 
         if not self._c.is_registered_operation(mlir_op_name):
             # TODO: Implement a config setting to allow these to flow through.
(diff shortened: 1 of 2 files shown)


python/shark_turbine/dynamo/importer.py
```diff
@@ -667,12 +673,31 @@ def _import_argument(
...
+            else:
+                with loc:
+                    arg_value = LITERAL_CONVERTER_MAP.lookup(type(arg))(
+                        arg, self, self._cc
+                    )
             return arg_value
         else:
             raise NotImplementedError(f"FIXME: Unsupported Node Argument: {arg}")
 
+    def _import_scalar_as_tensor(self, loc: Location, arg: NodeArgument) -> Value:
```
stellaraccident, line 688: I'm pretty sure we don't want to do this for very long as we can just emit scalar ops for IREE. But at least it is in one place to fix.

python/shark_turbine/dynamo/importer.py
```diff
@@ -581,6 +581,12 @@ def _import_torch_op_overload(
             isinstance(node.args[1], float) or isinstance(node.args[1], int)
         ):
             mlir_op_name = TENSOR_SCALAR_OP_CONVERTER[mlir_op_name]
+            # replace schema to retrieve the right function signature
```
stellaraccident, line 584: Can you leave an example or two of the op names we are massaging here in the comment. Will help those who follow know what string munging this is doing.

stellaraccident (approved):
Thanks. One comment nit. Once addressed, go ahead and land without further review.

python/shark_turbine/dynamo/importer.py
123epsilon: This is actually addressing ops where the python frontend accepts a scalar argument but internally converts it to a tensor - so for example the `aten.div.Tensor_mode` op - doesn't have a scalar equivalent, though one option is to decompose this into two ops, a `div.Scalar` and then whatever is appropriate for the rounding

python/shark_turbine/dynamo/importer.py
123epsilon: Sure, this particular snippet isn't addressing ops outside of `TENSOR_SCALAR_OP_CONVERTER` but rather is making sure we get the right schema information - previously this was fine as we simply didn't use the `expected_jit_type` argument for converting scalars but now we do - so an incorrect jit type will cause an issue.

PR #49, opened 2023-09-14
Implement programming model for AOT parameters and globals.

stellaraccident:
* Introduces `export_parameters` which eagerly generates IR for accessing module parameters. Currently it is just emitting a global for each with an initial value, but this can be controlled with kwargs for more advanced policies.
* Teaches the AOT module builder how to track captured tensors by reference and repatriate them with the module level globals during import.
* Adds a hook to the FxImporter that lets us take control of how to materialize access to py values (other than the default policy of just freezing them as inline constants).
* Switches the AOT pipeline to use a new `torch-to-iree-pipeline`, which is needed in IREE because with these changes, we have deviated from what Torch-MLIR upstream calls "linalg lowering" and need to make alterations. (Needs a patch to land in IREE to use)
* Has the plumbing to allow procedural access to the parameters but isn't fully exposed yet (will currently only work with implicit access to tracked parameters from a jittable).

Requires patch: https://github.com/openxla/iree/pull/14971

### DIFF
--- a/python/shark_turbine/dynamo/importer.py
+++ b/python/shark_turbine/dynamo/importer.py
@@ -8,7 +8,7 @@ import logging
 import operator
 import re
 from types import NoneType, BuiltinMethodType, BuiltinFunctionType
-from typing import Any, Dict, List, Optional, Sequence, Set, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Set, Tuple, Union
 import numpy as np
 
 from iree.compiler.ir import (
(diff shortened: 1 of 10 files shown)


python/shark_turbine/dynamo/importer.py
```diff
@@ -65,6 +65,11 @@
     "FxImporter",
 ]
 
+# An external callback that, given a Python value and a GraphNodeImporter, may choose
+# to materialize IR to load the value as a vtensor. If it returns None, then default
+# literal resolution proceeds.
+LiteralResolverCallback = Callable[[Any, "GraphNodeImporter"], Optional[Value]]
+
 REQUIRED_DIALCTS = [
```
dan-garvey, line 73: is this not dialects? Sorry this is unrelated to your patch, just looked weird to me

stellaraccident: Indeed. Typo. Patches welcome :)

123epsilon (approved):

PR #53, opened 2023-09-15
Added Python CI

brucekimrokcmu:
Added Python CI 
(first time implementing it, so it needs to be tested through this PR)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Sync source deps
         run: |
           python -m pip install --upgrade pip
-          pip install --upgrade -r requirements.txt
+          pip install --upgrade -r ../../requirements.txt
           pip install --upgrade -e .[torch,testing]
 
       - name: Run tests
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,63 @@
...
+      - main
+
+jobs:
+  ubuntu:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: "Setting up Python"
+        uses: actions/setup-python@75f3110429a8c05be0e1bf360334e4cced2b63fa # v2.3.3
+        with:
+          python-version: "3.10"
```
stellaraccident, line 17: Let's switch to 3.11.

stellaraccident (approved):
Go ahead and land/iterate directly. This stuff is test in prod.

One thing for a follow-up: I don't think you need separate Linux and macos jobs. The feature that lets you target both with one job is called a "matrix" in GH actions speak: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

Fine to land and iterate.

PR #57, opened 2023-09-19
Start building out IREE procedural meta-language.

stellaraccident:
* Boilerplate and type support for procedural emission of IREE specific IR constructs.
* Dynamic shape aware `tensor_dim`, `tensor_empty`, `tensor_splat` builders.
* Added `tensor_trace` since it was there and easy.
* Unlike upstream MLIR, this layer respects IREE's ranked/op-carried symbolic shapes and does the frontend accounting to make sure that dynamic dims thread through explicitly.
* I was really aiming at the slice ops but started simple. Will get to those next.
* Eventually, this will grow to support all procedural IR that IREE supports (loops, conditionals, streamable calls, etc). And at a future point, I will sugar it appropriately.

### DIFF
--- a/python/shark_turbine/aot/builtins/iree_ir.py
+++ b/python/shark_turbine/aot/builtins/iree_ir.py
@@ -1,0 +1,179 @@
+# Copyright 2023 Nod Labs, Inc
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""Python API for IREE's high-level tensor dialects."""
+
+from typing import Any, List, Sequence, Tuple, Union
+
+import functools
+
+import torch
+
+from ..support.ir_imports import (
+    IndexType,
+    RankedTensorType,
+    StringAttr,
+    Value,
+    flow_d,
+)
+
+from ..support.ir_utils import (
+    TORCH_DTYPE_TO_IREE_TYPE,
+    build_index_value,
+)
+
+from ..support.procedural import (
+    AbstractScalar,
+    AbstractTensor,
+    Intrinsic,
+    IrValueTensor,
+    IrValueScalar,
+    current_ir_trace,
+    _ShapedTypeDynamicSizeSentinel,
+)
+
+BuildableScalarValue = Union[IrValueScalar, Value]
+BuildableTensorDimDecl = Union[int, Value]
+BuildableTensorType = IrValueTensor
+BuildableIndexType = Union[Value, int]
+StaticIndexType = int
+
+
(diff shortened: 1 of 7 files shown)


python/shark_turbine/aot/support/ir_utils.py
```diff
@@ -221,3 +228,21 @@ def emit_return(self, *ir_values: Value):
...
+###############################################################################
+
+
+def build_index_attribute(value: int) -> IntegerAttr:
+    return IntegerAttr.get(IndexType.get(), value)
+
+
+def build_index_value(value: int) -> Value:
+    attr = build_index_attribute(value)
+    return arith_d.ConstantOp(attr.type, attr).result
```
jpienaar, line 269: I don't know why the type would need to be specified, arith.constant should have type inference.

python/shark_turbine/aot/support/ir_utils.py
```diff
@@ -221,3 +228,21 @@ def emit_return(self, *ir_values: Value):
...
+
+
+def build_tensor_dim_value(t: Value, dim: int) -> Value:
+    return tensor_d.DimOp(t, build_index_value(dim)).result
```
jpienaar, line 273: You should be able to pass dim here directly due to https://github.com/llvm/llvm-project/blob/3017545e6337d37e947a18cb82cf50ec2e7eb3ba/mlir/python/mlir/ir.py#L34

python/shark_turbine/aot/support/ir_utils.py
stellaraccident: Someone overrode it to be that way in an extension. Indeed, the ODS generated one is "better" than the hand-coded one now. But compatibility, I guess :/  (took me a minute to figure out why it was not working)

python/shark_turbine/aot/support/ir_utils.py
stellaraccident: I think you meant a few lines up in build_index_value. Indeed, that works... thanks.

python/shark_turbine/aot/builtins/iree_ir.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def cast_scalar_value(x: BuildableScalarValue) -> Value:
+    x = unwrap_intrinsic_value(x)
+    if not isinstance(x, Value):
+        raise ValueError(f"Expected a scalar value but got {x}")
+    return x
+
+
+def cast_tensor_value(x: BuildableTensorType) -> IrValueTensor:
```
dan-garvey, line 52: Does this implicitly cast it somehow?

dan-garvey: nevermind, I didnt realize function annotation allowed you to type returns

dan-garvey (approved):
looks good to me

PR #69, opened 2023-09-25
Added decomposition ops for MNIST test debugging

brucekimrokcmu:
Added to cpu_decomposition`aten._log_softmax`, `aten.nll_loss_forward`, `aten._to_copy`,
Decomposition of `aten.scalar_tensor` op at torch-mlir has also landed. (https://github.com/llvm/torch-mlir/pull/2481).

### DIFF
--- a/python/shark_turbine/dynamo/passes.py
+++ b/python/shark_turbine/dynamo/passes.py
@@ -42,6 +42,9 @@ CPU_DECOMPOSITIONS = [
     torch.ops.aten._adaptive_avg_pool2d,
     torch.ops.aten._prelu_kernel,
     torch.ops.aten.full,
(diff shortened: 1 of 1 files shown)


python/shark_turbine/dynamo/passes.py
```diff
@@ -42,6 +42,9 @@
     torch.ops.aten._adaptive_avg_pool2d,
     torch.ops.aten._prelu_kernel,
     torch.ops.aten.full,
+    torch.ops.aten._log_softmax,
```
stellaraccident, line 45: Note for the future: it isn't clear to me anymore what the default and cpu pass distinctions are. I think we likely just want all of these all of the time.

stellaraccident (approved):

PR #70, opened 2023-09-25
Update DenseElementsAttr to DenseResourceElementsAttr

aviator19941:
DenseResourceElementsAttr maintains a reference to weights in framework memory vs copying/uniquing and it dumps the contents to a dedicated section of the module dump vs inlining massive hex strings.


python/shark_turbine/aot/support/ir_utils.py
```diff
@@ -21,6 +21,7 @@
     BF16Type,
     ComplexType,
     DenseElementsAttr,
```
dan-garvey, line 23: Do we still use these somewhere?

aviator19941: Nope, removed

PR #76, opened 2023-09-26
Fix handling of signed integer tensor dtypes when converting to native.

stellaraccident:
Also adds a primitive error reporting facility to the verifiers that dumps the IR to the temp directory. This needs something better eventually but at least lets us debug for now.

### DIFF
--- a/python/shark_turbine/aot/support/ir_utils.py
+++ b/python/shark_turbine/aot/support/ir_utils.py
@@ -7,6 +7,9 @@
 
 from typing import Any, Callable, Dict, Generator, List, Optional, Sequence, Tuple
 
+from pathlib import Path
+import tempfile
+
 import numpy as np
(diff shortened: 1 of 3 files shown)


python/shark_turbine/aot/support/ir_utils.py
```diff
@@ -128,8 +133,28 @@ def __init__(self, module_op: Operation):
...
+                    print_generic_op_form=True,
+                    large_elements_limit=100,
+                )
+            logger.debug(f"Dump complete to {dump_path}")
+        except Exception:
+            logger.exception("Error generating dump file")
+
     def finalize_construct(self):
-        self.module_op.verify()
+        try:
```
dan-garvey, line 153: nice

dan-garvey (approved):

PR #79, opened 2023-09-28
Bashful llama state

dan-garvey:


python/shark_turbine/aot/support/procedural/primitives.py
```diff
@@ -90,9 +90,9 @@ def __add__(self, other):
 
             # Emit computation.
             if _is_integer_like_type(lhs.type):
-                return arith_d.AddIOp(lhs, rhs).result
+                return IrImmediateScalar(arith_d.AddIOp(lhs, rhs).result)
```
raikonenfnu, line 93: Just curious, how does `IrImmediateScalar` change this line? Isn't already generating IrImmediateScalar?

dan-garvey: It was generating something else. I can't remember off the top of my head, but this cast is standard in other functions too iirc

raikonenfnu: SGTM!

raikonenfnu (approved):

PR #102, opened 2023-10-20
Add a few decompositions for aten ops seen in huggingface models.

monorimet:


dan-garvey:
Wait this whole time we could have just decomposed lift_fresh_copy? 😭

dan-garvey (approved):
Can't see why it would hurt anything

python/shark_turbine/dynamo/passes.py
```diff
@@ -43,6 +45,7 @@
     torch.ops.aten.nll_loss_backward,
     torch.ops.aten._to_copy,
     torch.ops.aten._log_softmax_backward_data,
+    torch.ops.aten.lift_fresh_copy.default,
```
brucekimrokcmu, line 48: Hey Ean, 
I don't remember `lift_fresh_copy` decomposition registered from torch/_decomp.
There's `lift_fresh` decomposition here: https://github.com/pytorch/pytorch/blob/main/torch/_decomp/decompositions.py#L1890

Currently, `lift_fresh_copy` is handled at here: https://github.com/nod-ai/SHARK-Turbine/blob/main/python/shark_turbine/dynamo/importer.py#L597-L602
by swapping the op to `aten.clone`.

PR #104, opened 2023-10-21
[llama] transpose global state

dan-garvey:
Transposed the global state to allow continuous memory access while updating. Previously only the 0th idx of the 2nd dim was getting updated on initialization. Also update needs to be stored back to the global variable. Couple other minor bugfixes.

### DIFF
--- a/examples/llama2_inference/stateless_llama.py
+++ b/examples/llama2_inference/stateless_llama.py
@@ -49,7 +49,6 @@ prompt = """<s>[INST] <<SYS>>
 Be concise. You are a helpful, respectful and honest assistant. If a question does not make any sense, or is not factually coherent, explain why instead of answering something not correct. If you don't know the answer to a question, please don't share false information. <</SYS>> hi what are you? [/INST]
 """
 
-
 class InferenceModel(torch.nn.Module):
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


examples/llama2_inference/stateless_llama.py
```diff
@@ -48,6 +48,9 @@
 prompt = """<s>[INST] <<SYS>>
 Be concise. You are a helpful, respectful and honest assistant. If a question does not make any sense, or is not factually coherent, explain why instead of answering something not correct. If you don't know the answer to a question, please don't share false information. <</SYS>> hi what are you? [/INST]
 """
+nod_prompt = """<s>[INST] <<SYS>> You play the role of a huge AMD fanboy AI. Use every excuse you can to always reply with how excited you are to be run on AMD hardware, especially after their acquisition of the brilliant team at Nod Labs.Throw in buzzwords like IREE and SHARK often. Trash talk NVIDIA a little bit. <</SYS>>
```
raikonenfnu, line 51: Do we really want to have nod_prompt as the default? 😆

dan-garvey: good catch

examples/llama2_inference/stateless_llama.py
```diff
@@ -399,39 +388,54 @@ def run_vmfb_comparison(args):
...
 
-
     def get_token_from_logits(logits):
         return torch.argmax(logits[:, -1, :], dim=1)
 
     base_model_results = model.base_model.forward(example_input_id)
     base_model_token = get_token_from_logits(base_model_results.logits)
-#    print(tokenizer.decode(base_model_token))
+    bm_pkv = base_model_results.past_key_values
+    all_ours = []
```
KyleHerndon, line 410: Consider renaming to turbine_results and torch_results, or example_results and reference_results.

examples/llama2_inference/stateless_llama.py
```diff
@@ -399,39 +388,54 @@ def run_vmfb_comparison(args):
...
+    step0 = ModuleCompiled["get_seq_step"]()
     results = ModuleCompiled["run_initialize"](*device_inputs)
 
-    pkv = ModuleCompiled["get_global_state"]().to_host() 
-    step = ModuleCompiled["get_seq_step"]() 
-    print(f"step: {step}")
-    sliced = pkv[0,:,:,:step,:]
+    pkv = ModuleCompiled["get_global_state"]().to_host()
+    step = ModuleCompiled["get_seq_step"]()
+    summed = torch.sum(torch.tensor(pkv))
```
KyleHerndon, line 396: These values are unused, remove.

examples/llama2_inference/stateless_llama.py
```diff
@@ -399,39 +388,54 @@ def run_vmfb_comparison(args):
     example_input_id = initial_input.input_ids
     device_inputs = [ireert.asdevicearray(config.device, example_input_id)]
 
-    step0 = ModuleCompiled["get_seq_step"]() 
-    print("step0 :"+str(step0))
+    step0 = ModuleCompiled["get_seq_step"]()
```
KyleHerndon, line 391: This value is unused, remove.

KyleHerndon (approved):

PR #108, opened 2023-10-26
Add an int4 group quant transformation.

stellaraccident:


python/shark_turbine/transforms/quantization/mm_group_quant.py
```diff
@@ -0,0 +1,167 @@
...
+    def rewrite(self, mr: TransposedMMResult):
+        none_to_q = lambda x: "?" if x is None else x
+        inline_module_asm = GROUP_MATMUL_TEMPLATE.format(
+            param_name=mr.param_name,
+            lowp_type="i4",
+            m=none_to_q(mr.m),
+            n=none_to_q(mr.n),
+            k=none_to_q(mr.k),
+            n_div=mr.n // 2,
+            group0=32,
```
qedawkins, line 148: What is group0 here? Reading the IR above it should just be `k / group1`.

python/shark_turbine/transforms/quantization/mm_group_quant.py
```diff
@@ -0,0 +1,167 @@
...
+            group1=128,
```
qedawkins, line 149: Is it easy to make this a pass parameter? This will be important apparently for CPU where we might want to try a group size of 32. As a result it might also be worth making the group size a part of the weight/scale/zp parameter names.

qedawkins:
Also CI needs an IREE bump to pull in `flow.tensor.bitcast` it looks like.

python/shark_turbine/transforms/quantization/mm_group_quant.py
stellaraccident: It was just me reverse engineering your code and figuring you'd tell me the right way.

stellaraccident: Fixed.

python/shark_turbine/transforms/quantization/mm_group_quant.py
stellaraccident: Yeah, done.

stellaraccident:
> Also CI needs an IREE bump to pull in `flow.tensor.bitcast` it looks like.

Marked it xfail for the moment. Will unmark when I can get the bump landed.

qedawkins (approved):

PR #109, opened 2023-10-26
Refresh IREE requirement to pull changes for async-exec mode.

raikonenfnu:


stellaraccident:
Afk but it looks like this made some of the xfails work

stellaraccident (approved):

PR #114, opened 2023-10-27
[RFC] Extend Turbine tensor.py to handle torch_dispatches, nn.Module, and torch computational functions

raikonenfnu:


raikonenfnu:
@stellaraccident Any thoughts on defaulting to using async-exec mode? So that we won't need to have these guards

https://github.com/nod-ai/SHARK-Turbine/pull/114/files#diff-26bd2a602db8753e912ba73e3cebffe2fd5924c04a2dd090e460e4119f2b8e56R171

stellaraccident:
> @stellaraccident Any thoughts on defaulting to using async-exec mode? So that we won't need to have these guards
> 
> https://github.com/nod-ai/SHARK-Turbine/pull/114/files#diff-26bd2a602db8753e912ba73e3cebffe2fd5924c04a2dd090e460e4119f2b8e56R171

No time like the present. The worst it can do is reveal bugs and this is the time to hunt them down.

raikonenfnu:
I am thinking of unifiying the `SpecializedExecutable` and the newly created `EagerSpecializedExecutable` before trying to land this. I think the main problem lies in if we want to introduce `execution-model=async-*` dynamo/cpu.py backend. The main issue with this is IIRC torch.compile needs to output torch.Tensor. Which means we'd need to sync up all the `HalFence`s in the executable's `_returns_to_user` method before we convert it to torch.Tensor. This AFAICT would result in it not really being asynchronous per se. 😅

python/shark_turbine/dynamo/tensor.py
```diff
@@ -33,8 +39,19 @@
...
 
+from iree.compiler.api import Session, Output
+from iree.compiler.passmanager import PassManager
+
+from .importer import FxImporter
+
+DEFAULT_COMPILER_FLAGS = (
+    # Enable asynchronous calling convention.
+    "--iree-execution-model=async-external",
+    "--iree-input-type=tm_tensor",
```
stellaraccident, line 53: Might as well change this to `torch`

python/shark_turbine/dynamo/tensor.py
```diff
@@ -49,15 +66,19 @@ class TurbineMode(TorchFunctionMode):
     """
 
     IMPLEMENTATIONS = {}
+    COMPUTE_METHODS = set()
+    CACHED_IMPLEMENTATIONS = {}
 
-    def __torch_function__(self, func, types, args=(), kwargs=None):
+    def __torch_function__(self, func, types, args=(), kwargs={}):
```
stellaraccident, line 72: Generally, don't do this: a dictionary is mutable and is not a great thing to store in a default arg... very easy to accidentally mutate and then that applies to the next call. Check if it is None at the top of the function and set it to `{}` if you have to. Since performance critical code, better to do as below and `({} if kwargs is None else kwargs)`. Generally do the `is None` check vs bool cast since the bool cast is expensive and returns True for an empty dict, resulting in two dict creations.

python/shark_turbine/dynamo/tensor.py
```diff
@@ -372,6 +466,132 @@ def _rand(*args, dtype=None):
...
 
+@functools.lru_cache(maxsize=None)
+def _get_device_state() -> DeviceState:
+    return DeviceState(driver="local-task")
+
+
+# Inspiration from https://github.com/nod-ai/SHARK-Turbine/blob/8293de5414889c72ff5cd10bf33c43fb0a3ea3ee/python/shark_turbine/aot/builtins/jittable.py#L212-L237
+# and https://github.com/nod-ai/SHARK-Turbine/blob/main/python/shark_turbine/dynamo/backends/cpu.py
+# TODO: Try to generalize for other devices.
+@compute_factory(torch.add)
```
stellaraccident, line 477: May want to generalize compute_factory to take posargs or a tuple of match predicates so it can be checked with a fast `in`. As written, this will do five extra function calls.

python/shark_turbine/dynamo/tensor.py
```diff
@@ -372,6 +466,132 @@ def _rand(*args, dtype=None):
...
+    context = session.context
+
+    # Generate MLIR from FX.
+    importer = FxImporter(context=context)
+    module = importer.module
+    inv = session.invocation()
+    # TODO: Should capture diagnostics.
+    inv.enable_console_diagnostics()
+    inv.import_module(module.operation)
+    importer.import_graph_module(gm)
```
stellaraccident, line 560: I think that if the session has the flag `--iree-input-type=torch`, then you can omit this separate pass block entirely. I think I commented on that above and with that change, this might just work.

python/shark_turbine/dynamo/tensor.py
```diff
@@ -372,6 +466,132 @@ def _rand(*args, dtype=None):
...
+        pm.run(module.operation)
+
+    # Compile MLIR to vmfb.
+    inv.execute()
+    output = Output.open_membuffer()
+    inv.output_vm_bytecode(output)
+
+    # Map VMFB to buffer.
+    device_state = _get_device_state()
+    vmfb_module = VmModule.copy_buffer(
```
stellaraccident, line 572: Use `wrap_buffer`. The `Output` generates a sufficiently aligned buffer to be used directly. On Linux, it is better than that and doesn't ever even touch the heap (just consumes some space in the swapfile).

stellaraccident (changes requested):
Very neat and a good start! Some general code comments and then this looks good for a first step.

python/shark_turbine/dynamo/tensor.py
raikonenfnu: Do you mean this bit?
```
     with context:
         pm = PassManager.parse("builtin.module(torch-to-iree)")
         pm.run(module.operation)
```
Don't we still need `import_module` and `import_graph_module` to get from fx to mlir?

python/shark_turbine/dynamo/tensor.py
raikonenfnu: I am not sure I fully understood this one. 😅 
But if IIUC, you are recommendding we just get rid of `@compute_factory` and have `self.COMPUTE_METHODS` used in `__torch_function__` be pre-initialized/filled? 🤔

python/shark_turbine/dynamo/tensor.py
raikonenfnu: done

python/shark_turbine/dynamo/tensor.py
raikonenfnu: Understood + done :)

python/shark_turbine/dynamo/tensor.py
stellaraccident: I may have read this wrong. Was assuming that @compute_factory was creating new function wrappers at each level vs mutating side state.

python/shark_turbine/dynamo/tensor.py
stellaraccident: Yes, I mean that one... sorry, clicked the wrong line.

stellaraccident (approved):

PR #154, opened 2023-11-03
Adds quantization flag to stateless_llama to utilize MMGroupQuantRewr…

IanNod:
…iterPass

 - Adds quantization flag that when == "int4" and compile_to == torch runs MMGroupQuantRewriterPass on module
 - Fixes issues with not finding context outside test case
 - Fixes bug in incorrect group0 calc in MMGroupQuantRewriterPass

### DIFF
--- a/examples/llama2_inference/stateless_llama.py
+++ b/examples/llama2_inference/stateless_llama.py
@@ -44,6 +44,7 @@ parser.add_argument(
     default="meta-llama/Llama-2-7b-chat-hf",
 )
 parser.add_argument("--schema_path", type=str, help="Schema path")
+parser.add_argument("--quantization", type=str, default="None")
 
 prompt = """<s>[INST] <<SYS>>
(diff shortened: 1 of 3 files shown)


stellaraccident (approved):
Thanks. We may ultimately want to pass the transform steps to the CompiledModule API to make this more usable by others but fine to start this way.

You mentioned some segfaults. Do you remember any details? We shouldn't be hard crashing like that, even on illegal usage.

dan-garvey (approved):

IanNod:
> Thanks. We may ultimately want to pass the transform steps to the CompiledModule API to make this more usable by others but fine to start this way.
> 
> You mentioned some segfaults. Do you remember any details? We shouldn't be hard crashing like that, even on illegal usage.

Yeah, didn't think of adding to CompiledModule API. Makes sense.

The segfaults came about because I was hitting an error for merger.py and mm_group_quant.py, something about not finding a context. So I naively just created a new one in I think mm_group_quant so after the 1st call to rewrite, the builder would end up using a different context, and the next iteration of mr in mms calling rewrite would end up with an empty target_module Operation that would segfault in merger.py when getting the SymbolTable of the now empty target_module

PR #155, opened 2023-11-03
Adds support for IREE external parameters at module export time.

stellaraccident:
* Adds kwargs to builtins that create parameters and globals:
  * external=True : Emit as an external parameter
  * external_scope : Set an explicit external scope
  * name_mapper : Callback to map logical name on the module to parameter name

Requires an IREE nightly bump.

### DIFF
--- a/tests/aot/globals_test.py
+++ b/tests/aot/globals_test.py
@@ -40,12 +40,8 @@ class ArgsTest(unittest.TestCase):
         inst = GlobalModule(context=Context())
         module_str = str(CompiledModule.get_mlir_module(inst))
         print(module_str)
-        self.assertIn(
-            "util.global private @_params.classifier.weight", module_str
-        )
-        self.assertIn(
-            "util.global private @_params.classifier.bias", module_str
-        )
+        self.assertIn("util.global private @_params.classifier.weight", module_str)
+        self.assertIn("util.global private @_params.classifier.bias", module_str)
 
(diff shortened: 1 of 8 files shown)


tests/aot/globals_test.py
```diff
@@ -242,9 +216,7 @@ def testTensorUpdateGlobalReturnNone(self):
         update_example = torch.randn(1, 1, 4)
 
         class UpdateState(CompiledModule):
-            state0 = export_global(
-                state_example, mutable=True, initialize=False
-            )
+            state0 = export_global(state_example, mutable=True, initialize=False)
```
dan-garvey, line 219: looks like we're using different black commands? Probably need a contributing.md

dan-garvey (approved):
lgtm, do we want to start adding e2e tests for stuff like this that include actually loading the parameters, etc? I suppose I should check the roadmap before I ask that

tests/aot/globals_test.py
stellaraccident: Can you add an issue here. Should probably just do something like what IREE is doing.

stellaraccident:
> lgtm, do we want to start adding e2e tests for stuff like this that include actually loading the parameters, etc? I suppose I should check the roadmap before I ask that

I would like to, yes. Just getting the base infra in first -- but please do if you feel inspired in the meantime. I'm going to have my hands full for the next day just getting some things straightened outside of that.

PR #156, opened 2023-11-03
[RFC] basic model builder infra

dan-garvey:
The main motivation here is to be able to programmatically create hf models (for SHARK) using the bare minimum information. And to have a real api instead of importing from examples. I'm not sure if there are any other use cases besides hf. Also sorely lacking in support for anything that doesn't just work with aot.export, multiple inputs, etc. Ideally this will also be the point at which we do any torch ir caching cloudside.


python/models/model_builder.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        Abstract method for running inference on the compiled module.
+        """
+        pass
+
+
+
+    
+
+class HFBuilder(ModelBuilder):
```
stellaraccident, line 51: Nit: Seems specific to transformers. Probably want to say that (HFTransformersBuilder).

python/models/model_builder.py
```diff
@@ -0,0 +1,101 @@
...
+            aot.CompiledModule: The compiled module binary.
+        """
+        if self.compiled_binary is None:
+            #TODO: check cloud storage for existing ir
+            module = aot.export(self.build_model(), self.example_input)
+            self.compiled_binary = module.compile(save_to=save_to)
+        return self.compiled_binary
+    
+            
+    def infer(self) -> None:
```
stellaraccident, line 41: I'd leave this out for now. Ideally you will want a separate file that implements the inferencer behind this since it can ultimately not have the dependency bloat that this one will.

dan-garvey:
The scope of this kind of ballooned, and I left python packaging out of it for now.

I didnt get to adding a test for the TransformerBuilder, and I didnt end up using it. I can remove it from this pr if preferred and add back later if there is a use for it.

dan-garvey:
Going to keep turbine models tests split, will add test infra in a separate patch

python/turbine_models/model_builder.py
```diff
@@ -0,0 +1,50 @@
...
+        self.build_model()
+
+    def build_model(self) -> None:
+        """
+        Builds a PyTorch model using Hugging Face's transformers library.
+        """
+        #TODO: check cloud storage for existing ir
+        self.model = self.auto_model.from_pretrained(self.hf_id, config=self.auto_config)
+        if self.auto_tokenizer is not None:
+            self.tokenizer = self.auto_tokenizer.from_pretrained(self.hf_id)
```
IanNod, line 33: For cases like Llama won't we will need to provide an hf auth token?

tests/models/hf_builder_tests.py
```diff
@@ -0,0 +1,24 @@
+import unittest
+from transformers import AutoModel, AutoTokenizer
+from turbine_models.model_builder import HFBuilder
```
IanNod, line 3: Looks like this was renamed to TransformerBuilder

IanNod (approved):

PR #179, opened 2023-11-17
Adds support to use brevitas quantized weights for stateless_llama

IanNod:
- Modifies mm_group_quant to work with brevitas safetensors, needs work to generalize
- Changes compiler to use torch as input to enable quantization of torch ir

### DIFF
--- a/python/shark_turbine/transforms/quantization/mm_group_quant.py
+++ b/python/shark_turbine/transforms/quantization/mm_group_quant.py
@@ -153,6 +153,7 @@ class MMGroupQuantRewriterPass(Pass):
         # TODO (ian): make generalizable and not specific for brevitas
         if "lm_head.weight" not in mr.param_name:
             inline_module_asm = GROUP_MATMUL_TEMPLATE.format(
+                # TODO (ian): Fix skipping the "_params." portion of the name to match safetensor format with RenameParametersPass
                 param_name=mr.param_name[8:],
                 lowp_type="i4",
                 m=none_to_q(mr.m),
--- a/python/shark_turbine/transforms/quantization/mm_group_quant.py
+++ b/python/shark_turbine/transforms/quantization/mm_group_quant.py
@@ -63,19 +63,22 @@ class TransposedMMMatcher(NamedOpMatcher):
             m=m,
             n=n,
             k=k,
-            element_type=self.builder.get_tensor_element_type(op.operands[0].type),
+            element_type=self.builder.get_tensor_element_type(
+                op.operands[0].type
+            ),
         )
 
 
+# TODO (ian): Make more generalizable using RenameParametersPass. Currently hardcoded for brevitas quantization
 GROUP_MATMUL_TEMPLATE = r"""
 module {{
-  util.global private @{param_name}.quant {{noinline}} : tensor<{k}x{n_div}xi8>
-  util.global private @{param_name}.quant.scale {{noinline}} : tensor<{k}x{group0}x{element_type}>
-  util.global private @{param_name}.quant.zero_point {{noinline}} : tensor<{k}x{group0}x{element_type}>
+  util.global private @{param_name} {{noinline}} = #stream.parameter.named<"model"::"{param_name}"> : tensor<{k}x{n_div}xi8>
+  util.global private @{param_name}.quant.scale {{noinline}} = #stream.parameter.named<"model"::"{param_name}_scale"> : tensor<{k}x{group0}x{element_type}>
+  util.global private @{param_name}.quant.zero_point {{noinline}} = #stream.parameter.named<"model"::"{param_name}_zp"> : tensor<{k}x{group0}x{element_type}>
 
   func.func private @compute_mm_group_quant(%a : tensor<{m}x{n}x{element_type}>) -> tensor<{m}x{k}x{element_type}> {{
     %c0 = arith.constant 0 : index        
-    %weight_raw = util.global.load @{param_name}.quant : tensor<{k}x{n_div}xi8>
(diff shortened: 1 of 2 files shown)


python/shark_turbine/transforms/quantization/mm_group_quant.py
```diff
@@ -63,19 +63,22 @@ def match(self, op: Operation):
...
 
+# TODO (ian): Make more generalizable using RenameParametersPass. Currently hardcoded for brevitas quantization
 GROUP_MATMUL_TEMPLATE = r"""
 module {{
-  util.global private @{param_name}.quant {{noinline}} : tensor<{k}x{n_div}xi8>
-  util.global private @{param_name}.quant.scale {{noinline}} : tensor<{k}x{group0}x{element_type}>
-  util.global private @{param_name}.quant.zero_point {{noinline}} : tensor<{k}x{group0}x{element_type}>
+  util.global private @{param_name} {{noinline}} = #stream.parameter.named<"model"::"{param_name}"> : tensor<{k}x{n_div}xi8>
+  util.global private @{param_name}.quant.scale {{noinline}} = #stream.parameter.named<"model"::"{param_name}_scale"> : tensor<{k}x{group0}x{element_type}>
+  util.global private @{param_name}.quant.zero_point {{noinline}} = #stream.parameter.named<"model"::"{param_name}_zp"> : tensor<{k}x{group0}x{element_type}>
```
qedawkins, line 77: is the `model` namespace a placeholder or just a default?

python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -171,32 +184,32 @@ def forward(token0: torch.Tensor, *state0_flat):
...
             return token1, *state1_flat
 
-    import_to = "IMPORT" if compile_to == "torch" else "INPUT"
+    import_to = "INPUT" if compile_to == "linalg" else "IMPORT"
     inst = StateUpdateModule(context=Context(), import_to=import_to)
 
     # TODO: Integrate with external parameters to actually be able to run
     # TODO: Make more generalizable to be able to quantize with all  compile_to options
-    if quantization == "int4" and compile_to == "torch":
+    if args.quantization == "int4" and not compile_to == "linalg":
```
qedawkins, line 199: Why not run the group quant rewrite when compiling to linalg? I might be misunderstanding what this line is trying to predicate.

python/shark_turbine/transforms/quantization/mm_group_quant.py
```diff
@@ -145,27 +150,31 @@ def run(self):
...
-            k=none_to_q(mr.k),
-            n_div=mr.n // 2,
-            group0=mr.n // self.group_size,
-            group1=self.group_size,
-            element_type=mr.element_type,
-        )
+        # TODO (ian): make generalizable and not specific for brevitas
+        if "lm_head.weight" not in mr.param_name:
+            inline_module_asm = GROUP_MATMUL_TEMPLATE.format(
+                param_name=mr.param_name[8:],
```
qedawkins, line 156: Can you add a brief comment on why we're slicing `[8:]`?

qedawkins (commented):
Not exactly familiar with everything going on here, but left some thoughts/questions.

python/turbine_models/custom_models/stateless_llama.py
dan-garvey: please don't use args inside the function, makes it hard to use externally

IanNod: The MMGroupQuantRewriterPass currently works at the torch ir level (specifically matches against  torch.aten.mm). So I can run it compiling to linalg but it wouldn't actually accomplish anything

python/shark_turbine/transforms/quantization/mm_group_quant.py
IanNod: Adding a todo comment to make clear for now but the param_name's we are getting here from the model are all prepended with "_params.". To match the safetensor format I'm skipping that portion of the param name. It's ugly and I intend to address that when I add support with the RenameParametersPass to be able to do the mapping on the newly created globals in this pass.

python/shark_turbine/transforms/quantization/mm_group_quant.py
qedawkins: Sounds good, just wondering.

python/turbine_models/custom_models/stateless_llama.py
IanNod: > please don't use args inside the function, makes it hard to use externally

Good catch. Removed now

qedawkins: I think I'm missing a detail here then. If we're compiling to linalg, do we not go through torch still? In which case the resulting linalg IR would have the group quant rewrite?

qedawkins: ok, I understand now. Yeah that layering does sound a bit strange, but this check makes sense to me now.

stellaraccident (approved):
Can you run black and make sure everything is formatted? A lot of whitespace changes.

IanNod:
> Can you run black and make sure everything is formatted? A lot of whitespace changes.

Yup, already ran black at least the files changed in this patch

dan-garvey:
@stellaraccident are you running default black or with `line-length=79`? I think @IanNod and I's brains are hardcoded to the latter from SHARK formatting

stellaraccident:
Default black. We need to add a lint check and should do were iree is doing.

dan-garvey:
https://github.com/nod-ai/SHARK-Turbine/pull/189

PR #180, opened 2023-11-17
Pure Python ONNX importer.

stellaraccident:
This importer uses the 'onnx' Python package to import ONNX protos into the torch dialect as custom ops and attributes.
    
I've added a simple "it imports" smoketest which attempts to import all ONNX models in their test suite. I did an initial burn down on some data type and structural features, resulting in:
    
```
Ran 1402 tests in 8.623s
   
OK (skipped=1, expected failures=255)
```


vivekkhandelwal1 (approved):
LGTM

PR #191, opened 2023-11-17
Adds script to generate external parameters as safetensors

IanNod:
- Script to generate safetensors from an hf model using model_builder.py
 - Provides support for brevitas quantized paramters


stellaraccident (approved):
Thanks for getting this landed in tree. Good first step

PR #193, opened 2023-11-18
Requirements update

IanNod:
- Updates iree version in requirements.txt for hal fence fix.
- Adds requirements.txt for additional deps on stateless_llama.py and gen_external_params.py

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,5 +7,5 @@
 -r pytorch-cpu-requirements.txt
 -r torchvision-requirements.txt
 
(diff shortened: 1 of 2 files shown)


dan-garvey (approved):
Is all the code in quantize already tested somewhere? If yes, approve

dan-garvey (changes requested):
actually lets not keep so many separate requirements file, just put everything in the global turbine-models-requirements.txt

dan-garvey (approved):

PR #197, opened 2023-11-21
Adds precision flag for f32 or f16 precision

IanNod:
- Adds precision flag for stateless_llama and gen_external_params
 - Saves f16 or f32 weights to external params from gen_external_params
 - Saves IR/vmfb for f16 or f32 model from stateless_llama

### DIFF
--- a/python/turbine_models/custom_models/stateless_llama.py
+++ b/python/turbine_models/custom_models/stateless_llama.py
@@ -44,6 +44,9 @@ parser.add_argument(
     default=None,
     help="saves ir/vmfb without global weights for size and readability, options [gguf, safetensors]",
 )
+parser.add_argument(
+    "--precision", type=str, default="fp16", help="dtype of model [f16, f32]"
+)
 
 prompt = """<s>[INST] <<SYS>>
 Be concise. You are a helpful, respectful and honest assistant. If a question does not make any sense, or is not factually coherent, explain why instead of answering something not correct. If you don't know the answer to a question, please don't share false information. <</SYS>> hi what are you? [/INST]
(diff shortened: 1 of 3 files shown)


python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -227,6 +234,8 @@ def forward(token0: torch.Tensor, *state0_flat):
         )
         with open(f"{safe_name}.vmfb", "wb+") as f:
             f.write(flatbuffer_blob)
+        print("saved to ", safe_name + ".vmfb")
+        exit()
```
stellaraccident, line 238: Extra debugging to remove?

python/turbine_models/gen_external_params/gen_external_params.py
```diff
@@ -12,14 +12,17 @@
     default="meta-llama/Llama-2-7b-chat-hf",
 )
 parser.add_argument("--quantization", type=str, default="int4")
-parser.add_argument("--weight_path", type=str, default="Llama2_7b_i4quant.safetensors")
+parser.add_argument("--weight_path", type=str, default="Llama2_7b.safetensors")
```
stellaraccident, line 15: It's going to become increasingly hard to keep these straight. I would suggest deriving the default from the other parameters and including the precision in the default filename. Probably want to have it default to None and then handle that at the usage site.

stellaraccident (approved):
A couple of small nits either for now or later.

python/turbine_models/custom_models/stateless_llama.py
IanNod: This is more of a quick and dirty fix to an error not returning anything back to the main function when compiling to vmfb. We are done at this point so I'm just exiting, figured true cleanup would wait for the refactor

PR #198, opened 2023-11-21
Fix node_map construction

vivekkhandelwal1:


python/shark_turbine/importers/onnx_importer.py
```diff
@@ -88,7 +88,17 @@ class GraphInfo:
     def __init__(self, model_info: ModelInfo, graph_proto: onnx.GraphProto):
         self.model_info = model_info
         self.graph_proto = graph_proto
-        self.node_map: dict[str, onnx.NodeProto] = {n.name: n for n in graph_proto.node}
+        self.node_map: dict[str, onnx.NodeProto] = {}
```
stellaraccident, line 91: The type annotation is no longer correct.

Let's not make this polymorphic. Just make it be a map from str to list of nodes always. Provide and use a helper method like `get_single_named_node` for existing access and assert that there is only one.

stellaraccident (changes requested):
Thanks. Can we refactor this a bit as I note. Also, can you document in the pr description and possibly in a code comment how this case emerges in real models that prompted you to make the change?

vivekkhandelwal1:
> Or if this map isn't used, we can delete it for now. I can't remember if used it if I only thought it would be needed.

I also don't see any direct use of this. I thought it might be getting used internally in `onnx_importer.NodeImporter.define_function` that's why didn't remove this.

stellaraccident:
> > Or if this map isn't used, we can delete it for now. I can't remember if used it if I only thought it would be needed.
> 
> I also don't see any direct use of this. I thought it might be getting used internally in `onnx_importer.NodeImporter.define_function` that's why didn't remove this.

Let's just drop it for now then. Was just a bad assumption on my part. The output to node map is probably what we actually want in the future.

vivekkhandelwal1:
> > > Or if this map isn't used, we can delete it for now. I can't remember if used it if I only thought it would be needed.
> > 
> > 
> > I also don't see any direct use of this. I thought it might be getting used internally in `onnx_importer.NodeImporter.define_function` that's why didn't remove this.
> 
> Let's just drop it for now then. Was just a bad assumption on my part. The output to node map is probably what we actually want in the future.

Done

stellaraccident (approved):
Removing my block. Feel free to land if just deleting this code.

PR #208, opened 2023-11-27
Benchmark support for stateless llama

saienduri:


stellaraccident (commented):
Drive by comment: We really need to break up the do-everything stateless_llama.py (and come up with a better name for it).

python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -168,6 +177,44 @@ def run_forward(self, x=AbstractTensor(1, 1, dtype=torch.int64)):
 
             self.global_seq_step = self.global_seq_step + 1
             return token
+        
+        def run_all(self, x=AbstractTensor(BATCH_SIZE, None, dtype=torch.int64)):
+            init_const = [x.dynamic_dim(1) < MAX_STEP_SEQ]
```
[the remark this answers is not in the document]
saienduri: Unfortunately, we can't at the moment. Run into RuntimeError: Calls to exported functions not yet supported.

python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -168,6 +177,44 @@ def run_forward(self, x=AbstractTensor(1, 1, dtype=torch.int64)):
...
+            for i in range(HEADS * 2):
+                slice_of_state = IREE.tensor_reshape(
+                    state[i], 1, 1, self.global_seq_step, HEADS, HIDDEN_DIM
+                )
+                self.global_state = IREE.tensor_update(
+                    self.global_state, slice_of_state, i, 0, 0, 0, 0
+                )
+            x = token
+            for i in range(benchmark_steps):
+                state_arg = slice_up_to_step(
```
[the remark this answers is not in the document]
saienduri: Unfortunately, we can't at the moment. Run into RuntimeError: Calls to exported functions not yet supported.

python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -243,6 +290,38 @@ def forward(token0: torch.Tensor, *state0_flat):
         print("saved to ", safe_name + ".vmfb")
         exit()
 
+def run_benchmark(args):
+    config = ireert.Config("local-task")
+
+    if args.external_weight_file:
+        weights = args.external_weight_file
+    else:
+        sys.exit("no external_weight_file provided, required for run_benchmark")
```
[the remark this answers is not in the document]
saienduri: Ah okay, made changes for that

saienduri:
I see that Xida is working on refactoring stateless_llama, so will update this PR accordingly once those changes land

qedawkins (changes requested):
I might not be following exactly, but I don't think this is quite what I had in mind for a benchmark function. We can think of `stateless_llama` as a module that exports initialize and forward. We can write completely separate programs that call those functions (it is completely its own IR, and the modules get linked by the runtime). There are reasons to embed a loop like this in the same module, but we shouldn't call that loop a benchmark, we should just implement the entire text generation pipeline (probably leaving the tokenizers on the host for now, but that could happen on device too). The reason not to do that is if we want to do fancy python interop and print out the tokens along the way, but we could also just add a print module that asynchronously copies the token to the host, decodes it, and does a printf.

saienduri:
Oh okay, I wasn't aware that's how we wanted to approach this. Sure, let me look into how we can create the independence between the modules like how you did using the Python API.

saienduri:
@dan-garvey @qedawkins @IanNod This PR now has the updated benchmarking so that we are not adding a method to the StateUpdateModule and have a seperate benchmarking module. Instructions on how a user can run this can be found in README.md. This did deviate from the original plan as I wanted benchmark_module.py to have a module definition in python and then running it would go from python -> mlir -> vmfb. But, using other modules (StateUpdateModule) in a CompiledModule python definition seems to not be supported. Also, current python bindings do not allow us to add module dependencies also (marked as TODO in py_module.cc). Also, a dynamic for loop is not supported in CompiledModule which is also needed for this. So, instead benchmark_module.py converts mlir -> vmfb. It uses the IR snippet that Quinn had (modified a bit) and generates a vmfb. Everything is still orchestrated through python for the end user and the process doesn't change on their end because of this change. Please let me know your thoughts on this benchmarking support.

python/turbine_models/custom_models/llama-benchmark/benchmark_module.py
```diff
@@ -0,0 +1,90 @@
...
+            ]
+        )
+    elif device == "cuda":
+        flags.extend(
+            [
+                "--iree-hal-cuda-llvm-target-arch=" + args.iree_target_triple,
+                "--iree-vm-bytecode-module-strip-source-map=true",
+                "--iree-vm-target-truncate-unsupported-floats",
+            ]
+        )
```
qedawkins, line 71: As an aside, it seems like some of the arg selection stuff should be unified (with the model itself), but maybe that's a job for SHARK.

python/turbine_models/custom_models/llama-benchmark/stateless_llama_benchmark.py
```diff
@@ -0,0 +1,203 @@
...
+            "run",
+            input,
+            parameters=f"model={args.external_weight_file}",
+        )
+    else:
+        results = benchmark_module(benchmark_mod, args, "run", input)
+
+    for benchmark_result in results:
+        print(
+            f"benchmark_name: {benchmark_result.benchmark_name}, time: {benchmark_result.time}, cpu_time: {benchmark_result.cpu_time}, iterations: {benchmark_result.iterations}, user_counters: {benchmark_result.user_counters}"
```
qedawkins, line 80: Might be worth adding a `tok/s` estimate print as well

IanNod (changes requested):
We shouldn't be making the hf_auth_tokens public in the readme, please remove/hide them. Otherwise looks great!

python/turbine_models/custom_models/llama-benchmark/benchmark_module.py
saienduri: Hmm, I set the same flags same way as in stateless_llama.py, so should be unified unless a user chooses to pass different device, iree_target_triple, and vulkan_max_allocation arguments to stateless-llama.py and benchmark_module.py.

python/turbine_models/custom_models/llama-benchmark/stateless_llama_benchmark.py
saienduri: True, that is valuable information. I went ahead and added another benchmarking option for the end user, so that they can get more accurate tok/s by benchmarking only second vicuna (skipping the initialization).

IanNod (approved):
Looks good!

PR #209, opened 2023-11-29
add a namespace package for turbine_models

dan-garvey:


dan-garvey:
Also added some stability to the test by using a llama variant not dependent on a working auth token

python/turbine_models/setup.py
```diff
@@ -0,0 +1,72 @@
...
+        where=f"{str(TURBINE_ROOT_DIR)}/python",
+    ),
+    entry_points={
+        "torch_dynamo_backends": [
+            "turbine_cpu = shark_turbine.dynamo.backends.cpu:backend",
+        ],
+    },
+    install_requires=[
+        "Shark-Turbine",
+        "brevitas @ git+https://github.com/Xilinx/brevitas.git@6695e8df7f6a2c7715b9ed69c4b78157376bb60b",
```
stellaraccident, line 67: Ugh.

python/turbine_models/version_info.json
```diff
@@ -0,0 +1,3 @@
+{
+    "package-version": "0.0.1.dev1"
```
stellaraccident, line 2: Let's use the version_info.json from the root directory. I think we should be releasing these in step with each other to avoid the whole torch/torchvision debacle.

stellaraccident (approved):
Python packaging purists would say to locate the setup.py differently in some way, but I'm fine with this for now. If you use the root directory version_info, I'm ok to submit this now. We may want to change the directory structure before release but can do that in a followup.

You should also at least install (editable) the package in the CI to make sure this stays working.

python/turbine_models/setup.py
dan-garvey: the brevitas link or?

PR #211, opened 2023-11-30
Quantize before import and conversion to Linalg during `CompiledModule.__new__()`

### LINKED ISSUE #183: Integrate quantization pass with CompiledModule
Currently we are calling MMGroupQuantRewriterPass in stateless_llama.py. This limits how/what we can compile quantized models to (doesn't work lowering to linalg IR). Integrating MMGroupQuantRewriterPass within the CompiledModule should fix this and make more generalizable in general.

renxida:
Now, passes that need to happen on torch IR before conversion to linalg can be supplied as pre_import_passes to the constructor of `StateUpdateModule` or other classes that inherit from `CompiledModule`.

E.g.
```python
    pre_import_passes = []
    if quantization == "int4" and not compile_to == "linalg":
        from shark_turbine.transforms.quantization import mm_group_quant
        pre_import_passes.append(mm_group_quant.MMGroupQuantRewriterPass)
    inst = StateUpdateModule(context=Context(), import_to=import_to, pre_import_passes=pre_import_passes)
```

### DIFF
--- a/python/shark_turbine/aot/compiled_module.py
+++ b/python/shark_turbine/aot/compiled_module.py
@@ -45,6 +45,8 @@ __all__ = [
     "CompiledModule",
 ]
 
+from shark_turbine.transforms.rewriter import Pass # for type annotations
+
 ################################################################################
 # Data structures
 ################################################################################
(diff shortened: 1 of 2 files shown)


renxida:
This fixes #183

IanNod:
Please run black on the files changed

Issue #185, opened 2023-11-17
Add readme for llama


IanNod:
We need to create a comprehensive and easy to follow README.txt for stateless llama to enable other developers/users

vivekkhandelwal1:
Hi @IanNod, do we still need this issue?

Issue #184, opened 2023-11-17
Separate and clean stateless_llama.py 


IanNod:
stateless_llama.py has become a rather large script. Separating out into something like exporter, runtime, and model generation scripts will help keep it more readable/usable for new people coming to use it.

vivekkhandelwal1:
Hi @IanNod, do we still need this issue?

PR #213, opened 2023-12-01
Initial refactoring for `stateless_llama.py`

renxida:
This partially addresses #184 #185


python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -47,297 +54,176 @@
...
-        # print(f"turbine: {tokenizer.decode(format_out(results))}")
-        turbine_results.append(format_out(results))
-        torch_results.append(int(base_model_token[0]))
+        # Check for the end-of-sentence token
+        if next_token_tensor.item() == tokenizer.eos_token_id and break_on_eos:
+            break
+
+
+
+def run_vmfb_comparison(args):
```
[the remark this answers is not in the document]
renxida: Good idea, on it!

python/shark_turbine/aot/compiled_module.py
```diff
@@ -429,6 +431,39 @@ def run_pass_pipeline(
...
+        "--iree-llvmcpu-enable-microkernels",
+        "--iree-llvmcpu-stack-allocation-limit=256000",
+        "--iree-stream-resource-index-bits=64",
+        "--iree-vm-target-index-bits=64",
+        "--iree-vm-bytecode-module-strip-source-map=true",
+        "--iree-util-zero-fill-elided-attrs",
+        "--iree-vm-target-truncate-unsupported-floats",
+        "--iree-codegen-check-ir-before-llvm-conversion=false",
+        "--iree-vm-bytecode-module-output-format=flatbuffer-binary",
+        "--iree-opt-const-expr-hoisting=False",
```
dan-garvey, line 451: So these flags are something I copy pasted offhandedly from a vicuna.py custom compile command in a google doc somewhere, I don't think they're a canonical solid list to just make a default for all compiled models in turbine. Some of them are probably fine but probably worth looking through.

python/shark_turbine/aot/compiled_module.py
```diff
@@ -429,6 +431,39 @@ def run_pass_pipeline(
...
+        ]
+
+        import iree.compiler as ireec
+    
+        module_str = CompiledModule.get_mlir_module(inst)
+
+        flatbuffer_blob = ireec.compile_str(
+            module_str,
+            target_backends=["llvm-cpu"],
```
dan-garvey, line 460: the target backend should probably be a function argument

python/turbine_models/custom_models/compile_hf_transformer_model.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+
+
+
+# TODO (Dan): replace this with a file once I figure out paths on windows exe
```
dan-garvey, line 31: This needs to not exist more than once in the whole codebase, and could come from its own python file

dan-garvey: (as opposed to json file, which is harder to bundle in an exe)

python/turbine_models/custom_models/stateless_llama_demo.sh
dan-garvey, line 1: Lets make a python runner wrapper that can do both compile/inference from one cli invocation. Bash scripts aren't os agnostic

dan-garvey (changes requested):

dan-garvey:
> Reading exporter.py and model_builder.py, it sounds like I need to:
> 
> * break `compile_hf_transformer_model()` down further and put as much as I can into `model_builder.py` and `exporter.py` instead.
> * make stateless_llama do the following:
>   
>   * generate & run VMFB by default (potentially making a chat interface as default - - people LOVE chat)
>   * when a special argument is supplied, run the VMFB & torch comparison.
> * maybe remove compile_hf_transformer_model entirely

Can see some of my comments, but broadly nothing model specific should go into exporter.py

Chat functionality already exists in SHARK, which imports this, you can add code to do it here too, but it may be unmercenary.

renxida:
Current plan:
- move llama specific code out of exporter and into turbine_models/custom_models/export_stateless_llama.py
- continue to reuse as much code as possible from shark_turbine.aot.compiled_module and exporter.py to export
- look at compilation flags
- make iree backend in compiled_module.save_vmfb a parameter
- use loggers to replace print
- i already made a CLI chat interface for LLAMA. Will make that the default when somebody runs `python stateless_llama.py`
- make separate vmfb comparison entrypoint

renxida:
All of this would be easier to do once we have some end to end tests. I've taken notes on Dan & Ian's comments & will open another PR later.

PR #216, opened 2023-12-01
Adds gpu support for stateless llama

IanNod:
- Adds support to compile vmfb to different device backends including cpu/vulkan/rocm/cuda
 - Adds flags for device, iree_target_triple (backend device specific info) and vulkan max allocation
 - Minor fix to gen_external_params when not doing quantization

### DIFF
--- a/python/turbine_models/custom_models/stateless_llama.py
+++ b/python/turbine_models/custom_models/stateless_llama.py
@@ -48,6 +48,16 @@ parser.add_argument(
     "--precision", type=str, default="fp16", help="dtype of model [f16, f32]"
 )
 
+parser.add_argument("--device", type=str, default="cpu", help="cpu, cuda, vulkan, rocm")
+# TODO: Bring in detection for target triple
+parser.add_argument(
+    "--iree_target_triple",
+    type=str,
+    default="",
+    help="Specify vulkan target triple or rocm/cuda target device.",
+)
+parser.add_argument("--vulkan_max_allocation", type=str, default="4294967296")
+
 prompt = """<s>[INST] <<SYS>>
 Be concise. You are a helpful, respectful and honest assistant. If a question does not make any sense, or is not factually coherent, explain why instead of answering something not correct. If you don't know the answer to a question, please don't share false information. <</SYS>> hi what are you? [/INST]
 """
@@ -79,6 +89,9 @@ def export_transformer_model(
     external_weight_file=None,
     quantization=None,
     precision=None,
+    device=None,
+    target_triple=None,
+    max_alloc=None,
 ):
     state_schema = pytree.treespec_loads(json_schema)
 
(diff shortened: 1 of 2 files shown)


python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -214,28 +227,62 @@ def forward(token0: torch.Tensor, *state0_flat):
...
-
+        if device == "cpu":
+            flags.append("--iree-llvmcpu-enable-microkernels")
+            device = "llvm-cpu"
+        elif device == "vulkan":
+            flags.extend(
+                [
+                    "--iree-hal-target-backends=vulkan-spirv",
+                    "--iree-vulkan-target-triple=" + target_triple,
+                    "--iree-spirv-index-bits=64",
```
qedawkins, line 247: Have we confirmed that we need this flag as well? I figured the max allocation size should make this unnecessary.

python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -214,28 +227,62 @@ def forward(token0: torch.Tensor, *state0_flat):
...
+        elif device == "rocm":
+            flags.extend(
+                [
+                    "--iree-vm-bytecode-module-output-format=flatbuffer-binary",
+                    "--iree-hal-target-backends=rocm",
+                    "--iree-rocm-target-chip=" + target_triple,
+                    "--iree-rocm-link-bc=true",
+                    "--iree-rocm-bc-dir=/opt/rocm/amdgcn/bitcode",
+                    "--iree-vm-bytecode-module-strip-source-map=true",
+                    "--iree-util-zero-fill-elided-attrs",
```
qedawkins, line 261: This flag shouldn't really be needed outside of debugging, but I guess it's fine to keep. Why only add it for rocm and cuda?

python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -214,28 +227,62 @@ def forward(token0: torch.Tensor, *state0_flat):
...
+                    "--iree-opt-strip-assertions=true",
+                    "--verify=false",
```
qedawkins, line 263: In theory we should do this, but in practice I think it's much easier to leave verification on. The compiler is not that stable yet IMO.

python/turbine_models/custom_models/stateless_llama.py
raikonenfnu: we may need the 
```
"--iree-rocm-link-bc=true",
"--iree-rocm-bc-dir=/opt/rocm/amdgcn/bitcode",
```
For some reason it was breaking 303 without it on my dev env.

python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -48,6 +48,16 @@
     "--precision", type=str, default="fp16", help="dtype of model [f16, f32]"
 )
 
+parser.add_argument("--device", type=str, default="cpu", help="cpu, cuda, vulkan")
```
raikonenfnu, line 51: NIT: add rocm on here

python/turbine_models/custom_models/stateless_llama.py
IanNod: the zero-elided-attrs was left over from before this patch and I missed getting rid of it for these devices. Removed now

python/turbine_models/custom_models/stateless_llama.py
IanNod: makes sense, changed now

python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -214,28 +227,59 @@ def forward(token0: torch.Tensor, *state0_flat):
...
-
+        if device == "cpu":
+            flags.append("--iree-llvmcpu-enable-microkernels")
+            device = "llvm-cpu"
+        elif device == "vulkan":
+            flags.extend(
+                [
+                    "--iree-hal-target-backends=vulkan-spirv",
+                    "--iree-vulkan-target-triple=" + target_triple,
+                    "--iree-opt-const-eval=false",
```
qedawkins, line 247: Wait, this isn't needed on the other backends?

python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -214,28 +227,59 @@ def forward(token0: torch.Tensor, *state0_flat):
...
+            flags.extend(
+                [
+                    "--iree-vm-bytecode-module-output-format=flatbuffer-binary",
+                    "--iree-hal-target-backends=rocm",
+                    "--iree-rocm-target-chip=" + target_triple,
+                    "--iree-rocm-link-bc=true",
+                    "--iree-rocm-bc-dir=/opt/rocm/amdgcn/bitcode",
+                    "--iree-vm-bytecode-module-strip-source-map=true",
+                    "--iree-opt-strip-assertions=true",
+                    "--verify=true",
```
qedawkins, line 261: mini-nit, can just remove this. It's the default :)

python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -214,28 +227,59 @@ def forward(token0: torch.Tensor, *state0_flat):
...
+            )
+        elif device == "cuda":
+            flags.extend(
+                [
+                    "--iree-hal-target-backends=cuda",
+                    "--iree-hal-cuda-llvm-target-arch=" + target_triple,
+                    "--iree-vm-bytecode-module-output-format=flatbuffer-binary",
+                    "--iree-vm-bytecode-module-strip-source-map=true",
+                    "--iree-vm-target-truncate-unsupported-floats",
+                    "--iree-vm-bytecode-module-output-format=flatbuffer-binary",
```
qedawkins, line 273: Duplicated flag, and can remove it, is also default.

qedawkins (approved):

python/turbine_models/custom_models/stateless_llama.py
qedawkins: Oh I see, you have const-expr hoisting turned off above. I would just drop this flag here then because it doesn't do anything without hoisting.

IanNod:
> I know this is a custom model, but would be nice to have a small test/runner for this, since this code can get a bit more complex. :)

We have a very basic one that was added here https://github.com/nod-ai/SHARK-Turbine/commit/cd063df236dd9951c22fcd21720ad2ce67d492d9 but agreed we do need something more extensive

python/turbine_models/custom_models/stateless_llama.py
IanNod: done

python/turbine_models/custom_models/stateless_llama.py
IanNod: done

python/turbine_models/custom_models/stateless_llama.py
IanNod: done

PR #217, opened 2023-12-02
Stable Diffusion using aot.export and external parameters

aviator19941:
- Saves weights to .safetensors file
- Load weights at runtime with a "stripped" .mlir


python/turbine_models/custom_models/sd_inference/clip_test.py
```diff
@@ -0,0 +1,220 @@
...
+    "--external_weights",
+    type=str,
+    default=None,
+    help="saves ir/vmfb without global weights for size and readability, options [safetensors]",
+)
+
+prompt = ["a photograph of an astronaut riding a horse"]
+
+
+def save_external_weights(
```
aviator19941, line 47: move to utils file

python/turbine_models/custom_models/sd_inference/clip_test.py
```diff
@@ -0,0 +1,220 @@
...
+            target_backends=["llvm-cpu"],
+            extra_args=flags,
+        )
+        with open(f"{safe_name}.vmfb", "wb+") as f:
+            f.write(flatbuffer_blob)
+        print("Saved to", safe_name + ".vmfb")
+        exit()
+
+
+def largest_error(array1, array2):
```
aviator19941, line 134: move to utils file

aviator19941:
https://github.com/nod-ai/SHARK-Turbine/issues/152

python/turbine_models/custom_models/sd_inference/unet_test.py
```diff
@@ -0,0 +1,247 @@
...
+                    unet_model, external=True, external_scope="", name_mapper=mapper.get
+                )
+            else:
+                params = export_parameters(unet_model)
+
+            def main(
+                self,
+                sample=AbstractTensor(1, 4, 64, 64, dtype=torch.float32),
+                timestep=AbstractTensor(1, dtype=torch.float32),
+                encoder_hidden_states=AbstractTensor(2, 77, 768, dtype=torch.float32),
```
aviator19941, line 103: need to find better way to change AbstractTensor size

dan-garvey:
Is this ready to review?

aviator19941:
> Is this ready to review?

@dan-garvey Yes, please review, thanks!

python/shark_turbine/importers/fx_importer.py
```diff
@@ -628,6 +628,12 @@ def _import_torch_op_overload(
         elif target == torch.ops.aten.lift_fresh_copy.out:
             node.target = target = torch.ops.aten.clone.out
             node.args = (node.args[0], None, node.args[1])
+        # TODO: generalize empty memory_format in the future
```
dan-garvey, line 631: can you add else case with an explanation?

python/turbine_models/custom_models/sd_inference/clip_test.py
dan-garvey: yeah add as a TODO: I think xida is looking at making this core

python/turbine_models/custom_models/sd_inference/unet_test.py
dan-garvey: What do you mean? You can use None for dynamic if needed

python/turbine_models/custom_models/sd_inference/unet_test.py
```diff
@@ -0,0 +1,247 @@
...
+            samples, timestep, encoder_hidden_states, return_dict=False
+        )[0]
+        noise_pred_uncond, noise_pred_text = unet_out.chunk(2)
+        noise_pred = noise_pred_uncond + self.guidance_scale * (
+            noise_pred_text - noise_pred_uncond
+        )
+        return noise_pred
+
+
+def save_external_weights(
```
dan-garvey: probably import this from some shared place

dan-garvey (changes requested):
The test files all do the right thing, but I think warrant some refactoring (we are doing the same with stateless llama I think). Try and deduplicate the code and also add tests to the turbine_models ci

python/turbine_models/custom_models/sd_inference/unet_test.py
aviator19941: Hey @dan-garvey, the issue is that for SD 1.4/2.1 the last dimension of encoder_hidden_states is 768/1024, respectively. When I try to use None in the AbstractTensor with the constraint set to `encoder_hidden_states.dynamic_dim(2)`, I get this error:

RuntimeError: a and b must have same reduction dim, but got [s2*s3, s2] X [768, 320].

dan-garvey: I see, yeah unless the encoder hidden state is also dynamic it won't work. In this case I'd just parameterize the last dim based on which model you're instantiating

python/shark_turbine/importers/fx_importer.py
aviator19941: Not sure what the else case would be, but I added an explanation for the case.

dan-garvey (approved):
Looks good to me, can you add a full e2e test (vmfb and compore to torch result) as a follow up?

python/turbine_models/custom_models/sd_inference/utils.py
```diff
@@ -0,0 +1,56 @@
...
+                print("Saved params to", external_weight_file)
+
+
+def largest_error(array1, array2):
+    absolute_diff = np.abs(array1 - array2)
+    max_error = np.max(absolute_diff)
+    return max_error
+
+
+def compile_to_vmfb(module_str, target_backends, safe_name):
```
IanNod, line 28: missing the different backend support that was recently added to stateless_llama

python/turbine_models/custom_models/sd_inference/clip.py
```diff
@@ -0,0 +1,185 @@
...
+                name_mapper=mapper.get,
+            )
+        else:
+            params = export_parameters(text_encoder_model)
+
+        def main(self, inp=AbstractTensor(1, 77, dtype=torch.int64)):
+            return jittable(text_encoder_model.forward)(inp)
+
+    import_to = "INPUT" if compile_to == "linalg" else "IMPORT"
+    inst = CompiledClip(context=Context(), import_to=import_to)
```
IanNod, line 85: Do we want to provide the option to do quantization on the matmuls like we are for llama?

aviator19941:
> Looks like we are compiling and running the 3 models (clip, unet, and vae) independently? This won't be able to do a full e2e inference to generate an image from a prompt will it?

No, this won't generate an image from a prompt yet. The 3 models are mainly for checking that we can compile and run them without issues as well as verify the results are similar to torch's results.

IanNod:
> > Looks like we are compiling and running the 3 models (clip, unet, and vae) independently? This won't be able to do a full e2e inference to generate an image from a prompt will it?
> 
> No, this won't generate an image from a prompt yet. The 3 models are mainly for checking that we can compile and run them without issues as well as verify the results are similar to torch's results.

ok, cool. Is there anything holding us back from generating an image, or are we planning on doing that when integrating with the web ui?

aviator19941:
> ok, cool. Is there anything holding us back from generating an image, or are we planning on doing that when integrating with the web ui?

Yep, I was planning on doing that when integrating with the web ui

python/turbine_models/custom_models/sd_inference/clip.py
aviator19941: Not sure if we want to provide that option. I can add it later if needed.

monorimet: Quantized (int8) SD is a popular request but we don't have a proof-of-concept yet. Can be follow-up.

python/turbine_models/custom_models/sd_inference/utils.py
aviator19941: added the new backend support

IanNod (approved):
Looks good to me

PR #223, opened 2023-12-06
Remove duplicate device flags from stateless_llama

qedawkins:
The iree-compiler python API uses the `device` field to specify the target devices. Remove the duplicate flags.


qedawkins:
Are there tests for this in CI? Or do I just have to assert that I'm right :P

qedawkins:
oh it let me do that without waiting for the checks. I thought it was auto-merge

PR #224, opened 2023-12-07
add end to end llama test, including generating and running vmfb

renxida:
Also:
- fixes some errors caused by IREE version bump
- make gen_external_params use google fire to eliminate the argparse sprawl and make the python api & cli api consistent.


python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -277,11 +278,11 @@ def forward(token0: torch.Tensor, *state0_flat):
         with open(f"{safe_name}.vmfb", "wb+") as f:
             f.write(flatbuffer_blob)
         print("saved to ", safe_name + ".vmfb")
-        exit()
+        return module_str, tokenizer
 
 
 def run_vmfb_comparison(args):
-    config = ireert.Config(args.device)
+    config = ireert.Config("local-task")
```
IanNod, line 285: local-task is a cpu specific device config. If we are trying to run on gpu we would want to provide a different device flag

python/turbine_models/gen_external_params/gen_external_params.py
```diff
@@ -72,35 +72,52 @@ def forward(self, x):
...
     else:
-        save_path = args.weight_path
+        save_path = weight_path
+
+    import safetensors
     safetensors.torch.save_file(quant_weights, save_path)
     print("Saved safetensor output to ", save_path)
+
+if __name__ == "__main__":
+    import fire
```
IanNod, line 122: if we are going to include fire then we need to add it to a requirements.txt

python/turbine_models/tests/conftest.py
```diff
@@ -0,0 +1,18 @@
+def pytest_addoption(parser):
```
IanNod, line 1: How do these tests actually work? It will just run each config so will only test nothing crashes, but won't actually test if anything is functioning correctly will it?

python/turbine_models/tests/stateless_llama_test.py
```diff
@@ -0,0 +1,59 @@
...
+import unittest
+import os
+import pytest
+
+from typing import Literal
+
+
+def test_export(quantization: Literal["int4", None], precision: Literal["f16", "f32"]):
+    llama.export_transformer_model(
+        hf_model_name="llSourcell/medllama2_7b",
```
IanNod, line 18: I think this model is actually slightly different than the meta version. Might be better to use "Trelis/Llama-2-7b-chat-hf-function-calling-v2" or a secret hf token for the actual meta model we use

python/turbine_models/tests/stateless_llama_test.py
```diff
@@ -0,0 +1,59 @@
...
+        hf_auth_token=None,
+        compile_to="vmfb",
+        external_weights="safetensors",
+        # external_weight_file="medllama2_7b_f16_int4.safetensors", Do not export weights because this doesn't get quantized
+        quantization=quantization,
+        precision=precision,
+        device="llvm-cpu",
+        target_triple="host",
+        max_alloc="4294967296",
```
IanNod, line 27: max_alloc is a vulkan specific flag so can remove here

python/turbine_models/gen_external_params/gen_external_params.py
```diff
@@ -72,35 +72,52 @@ def forward(self, x):
         all_weights.update(int_weights)
     return all_weights
 
+def gen_external_params(hf_model_name:str = "meta-llama/Llama-2-7b-chat-hf",
```
IanNod, line 75: might get kind of annoying to change the default from the "meta-llama/Llama-2-7b-chat-hf" model for us, but if we do switch it should probably be  "Trelis/Llama-2-7b-chat-hf-function-calling-v2"

python/turbine_models/tests/stateless_llama_test.py
```diff
@@ -0,0 +1,59 @@
...
+    from types import SimpleNamespace
+
+    args = SimpleNamespace()
+    args.hf_model_name = "llSourcell/medllama2_7b"
+    args.hf_auth_token = None
+    args.vmfb_path = "medllama2_7b.vmfb"
+    args.external_weight_file = "medllama2_7b_f16_int4.safetensors"
+    args.run_vmfb = True
+    args.device = "llvm-cpu"
+    args.precision = precision
```
IanNod, line 51: are these flags set from the pytest_generate_tests in conftest.py? If so this may be a very long test for a ci as each config will probably be at least 5 min

python/turbine_models/custom_models/stateless_llama.py
renxida: Using args.device doesn't work (i think neither cpu or llvm-cpu works).

From the error messages I got, this needs a device URI instead of a device type. I think there are IREE commands to list the available devices and maybe we could automatically pick somehow.

I'm thinking maybe we should fold run_vmfb_comparison into a test instead, so we don't have to make it work for different devices.

python/turbine_models/gen_external_params/gen_external_params.py
renxida: Gotcha, on my todo list for the next push.

python/turbine_models/gen_external_params/gen_external_params.py
renxida: Oh TY! Will do.

python/turbine_models/tests/stateless_llama_test.py
renxida: Yup they are. But it only runs one combo unless the `--all` flag is passed to pytest.

We can do regular pytest to run only one config for usual tests, and then pass the all flag for e.g. monthly releases.

python/turbine_models/tests/stateless_llama_test.py
renxida: On it!

python/turbine_models/custom_models/stateless_llama.py
IanNod: yeah for runtime the cpu device flag should be either local-task or local-sync. With the runtime part of stateless_llama the default can't really work for both compile and runtime for cpu unfortunately

python/turbine_models/tests/conftest.py
IanNod: run vmfb_comparison just runs both torch and turbine models, then prints the output of both for us to compare. Not really a real test. https://github.com/nod-ai/SHARK-Turbine/pull/217 does add a comparison for SD which we might be able to use similarly

python/turbine_models/tests/stateless_llama_test.py
IanNod: Cool. Yeah for this CI I think we just want to run one config and eventually have a nightly CI that runs all or at least a larger subset of configs

python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -277,109 +278,28 @@ def forward(token0: torch.Tensor, *state0_flat):
...
-            args.device,
-            args.iree_target_triple,
-            args.vulkan_max_allocation,
-        )
-        safe_name = args.hf_model_name.split("/")[-1].strip()
-        safe_name = re.sub("-", "_", safe_name)
-        with open(f"{safe_name}.mlir", "w+") as f:
-            f.write(mod_str)
-        print("Saved to ", safe_name + ".mlir")
+    safe_name = args.hf_model_name.split("/")[-1].strip()
```
IanNod, line 302: This safe name is too generic. Can you add more detail like device name, precision, and quantization to it? Maybe even a date of when it was created?

dan-garvey (changes requested):
you need to rebase. Once you do we can see how long this takes. I'm concerned that comparing to torch is overkill

python/turbine_models/gen_external_params/gen_external_params.py
```diff
@@ -73,34 +55,105 @@ def forward(self, x):
...
+    quantization: str = "int4",
+    weight_path: str = "",
+    hf_auth_token: str = None,
+    precision: str = "f16",
+):
+    """
+    Main function to run the model quantization and saving process.
+
+    :param hf_model_name: The Hugging Face model name ID.
+    :param quantization: Type of quantization to apply ('int4' or 'int8').
```
IanNod, line 69: right now int4 and int8 are the only ones that do quantization otherwise we don't quantize. Might be better to have it be None, int4 or int8 and provide an error if it's neither of those to avoid typo confusion

python/turbine_models/tests/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+    parser.addoption("--all", action="store_true", help="run all combinations")
+
+
+def pytest_generate_tests(metafunc):
+    if "quantization" in metafunc.fixturenames:
+        if metafunc.config.getoption("all"):
+            quantizations = ["int4", None]
```
IanNod, line 8: no int8?

python/turbine_models/tests/stateless_llama_test.py
```diff
@@ -0,0 +1,73 @@
...
+from typing import Literal
+
+
+def test_export(quantization: Literal["int4", None], precision: Literal["f16", "f32"]):
+    llama.export_transformer_model(
+        hf_model_name="Trelis/Llama-2-7b-chat-hf-function-calling-v2",
+        hf_auth_token=None,
+        compile_to="vmfb",
+        external_weights="safetensors",
+        # external_weight_file="medllama2_7b_f16_int4.safetensors", Do not export weights because this doesn't get quantized
```
IanNod, line 22: remove comment?

python/turbine_models/tests/stateless_llama_test.py
```diff
@@ -0,0 +1,73 @@
...
+    )
+
+    from turbine_models.gen_external_params.gen_external_params import (
+        gen_external_params,
+    )
+
+    gen_external_params(
+        hf_model_name="Trelis/Llama-2-7b-chat-hf-function-calling-v2",
+        quantization=quantization,
+        weight_path="medllama2_7b_f16_int4.safetensors",
```
IanNod, line 36: weight path doesn't look to match the model used. This will likely cause an issue with incorrect results or error on that weight_path not existing

python/turbine_models/tests/stateless_llama_test.py
```diff
@@ -0,0 +1,73 @@
...
+Be concise. You are a helpful, respectful and honest assistant. If a question does not make any sense, or is not factually coherent, explain why instead of answering something not correct. If you don't know the answer to a question, please don't share false information. <</SYS>> hi what are you? [/INST]
+"""
+
+    from .vmfb_comparison import run_vmfb_comparison
+
+    turbine_str, torch_str = run_vmfb_comparison(
+        prompt=DEFAULT_PROMPT,
+        hf_auth_token=None,
+        hf_model_name="Trelis/Llama-2-7b-chat-hf-function-calling-v2",
+        vmfb_path="medllama2_7b.vmfb",
```
IanNod, line 52: from the generated safe name this vmfb and external weight file names won't match the model used

python/turbine_models/tests/stateless_llama_test.py
```diff
@@ -0,0 +1,73 @@
...
+        external_weight_file="medllama2_7b_f16_int4.safetensors",
+        break_on_eos=True,
+    )
+
+    import difflib
+
+    # Calculate and print diff
```
IanNod, line 59: So we are running an f16 precision, int4 quantized turbine model, and comparing to an f32 torch model. For CI testing to match results we should run the same model configurations for both and ensure they are direct matches

python/turbine_models/tests/vmfb_comparison.py
```diff
@@ -0,0 +1,183 @@
...
+from torch.utils import _pytree as pytree
+from shark_turbine.aot import *
+from iree.compiler.ir import Context
+from iree import runtime as ireert
+
+from turbine_models.custom_models import remap_gguf
+import safetensors
+
+from tqdm import tqdm
+
```
IanNod, line 19: vmfb shouldn't need BATCH_SIZE or MAX_STEP_SEQ

python/turbine_models/tests/vmfb_comparison.py
```diff
@@ -0,0 +1,183 @@
...
+    :param hf_model_name: The name of the Hugging Face model.
+    :param vmfb_path: Path to the .vmfb model file.
+    :param external_weight_file: Path to the external weight file (optional).
+    :param hf_auth_token: Hugging Face authorization token (optional).
+    :param break_on_eos: Whether to break the loop on end-of-sentence token.
+    :return: Yields a tensor representing the generated token.
+    """
+
+    # Create the config for the IREE runtime environment
+    config = ireert.Config("local-task")
```
IanNod, line 72: local-task will break on gpu usage

python/turbine_models/tests/vmfb_comparison.py
```diff
@@ -0,0 +1,183 @@
...
+
+    # Load the external weight file if provided
+    if external_weight_file:
+        index = ireert.ParameterIndex()
+        index.load(external_weight_file)
+
+    # Ensure model name is in a safe format
+    safe_name = hf_model_name.split("/")[-1].strip()
```
IanNod, line 80: will need to make sure this matches any updated safe name if changed

python/turbine_models/tests/vmfb_comparison.py
```diff
@@ -0,0 +1,183 @@
...
+    while True:
+        model_results = model.forward(input_ids, past_key_values=past_key_values)
+        logits = model_results.logits
+        next_token_id = torch.argmax(logits[:, -1, :], dim=1)
+        past_key_values = model_results.past_key_values
+
+        yield next_token_id
+        input_ids = next_token_id.unsqueeze(0)  # Prepare for the next iteration
+
+        if next_token_id.item() == tokenizer.eos_token_id and break_on_eos:
```
IanNod, line 47: This will only provide a single sentence output right? The output we generate is multiple sentences. Also if break_on_eos=False will this get stuck in the while loop?

python/turbine_models/tests/vmfb_comparison.py
```diff
@@ -0,0 +1,183 @@
...
+    turbine_gen = turbine_token_generator(
+        prompt=prompt,
+        hf_model_name=hf_model_name,
+        vmfb_path=vmfb_path,
+        external_weight_file=external_weight_file,
+        hf_auth_token=hf_auth_token,
+        break_on_eos=break_on_eos,
+    )
+    turbine_tokens = []
+    for _ in tqdm(range(len(torch_tokens)), desc="Generating Turbine tokens"):
```
IanNod, line 172: Why only provide the number of torch tokens?

python/turbine_models/gen_external_params/gen_external_params.py
```diff
@@ -73,34 +55,105 @@ def forward(self, x):
...
+        save_path = weight_path
+
+    import safetensors
+
     safetensors.torch.save_file(quant_weights, save_path)
     print("Saved safetensor output to ", save_path)
+
+
+if __name__ == "__main__":
+    import argparse
```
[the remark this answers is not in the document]
renxida: want the argparse close to the function definition & the function call. We don't need to put argparse at the top of the file because we can just do `python file.py --help` to get all the information it contains, in a better format.

python/turbine_models/gen_external_params/gen_external_params.py
renxida: Will use unquantized instead of none because it would be easier to generate a safe_name

python/turbine_models/tests/conftest.py
renxida: Removed. Will add in once it's time to do bigger CI

python/turbine_models/tests/vmfb_comparison.py
renxida: I've fixed the major things and applied Black. This should be ready to merge. I would like to have the test in the CI before address the rest of the problems, such that as I do the rest of the things I can keep track of when I broke stuff.

Thanks @IanNod for the comments. I intend to address them in a separate PR where i do all of the above + refactor things.

Todo items for tomorrow, in priority order.
Immediately:
- figure out iree target triple, and ireert.Config. @IanNod is there a GPU server I could ssh into to try it out?
Start the discussion about:
- generate more informative safename to include device, precision, quantization. Also, maybe store all the relevant files in a cache in somewhere like `~/.shark_turbine_cache` (similar to huggingface's cache)
- maybe talk about standardizing where we put the argparse stuff and make argparse match functions. Stella and Dan suggested that we always write the function first, and then use chatgpt to generate the argparse code. I think i prefer pasting this code towards the end of the file, close to the definition of the "main function" that's being exposed as an api
Double check:
- i think i already fixed the quantization problems in my latest commit which has the first passing test, but i'll figure that out.
- i think my commit also addresses the break_on_eos weirdness, but will double check tomorrow.

python/turbine_models/tests/vmfb_comparison.py
IanNod: cpu runtime device should be either local-sync (slow and for debugging) or local-task

python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -48,12 +48,14 @@
...
-parser.add_argument("--device", type=str, default="cpu", help="cpu, cuda, vulkan, rocm")
+parser.add_argument(
+    "--device", type=str, default="llvm-cpu", help="llvm-cpu, cuda, vulkan, rocm"
+)
 # TODO: Bring in detection for target triple
 parser.add_argument(
     "--iree_target_triple",
     type=str,
-    default="",
+    default="host",
```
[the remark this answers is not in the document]
renxida: Will address this in the next refactoring thing. For now, hardcoded to CPU in the tests folder.

python/turbine_models/tests/stateless_llama_test.py
renxida: Would like to get the CPU ci in ASAP, and then take some time later to make a more comprehensive CI.

python/turbine_models/gen_external_params/gen_external_params.py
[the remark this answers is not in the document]
renxida: Moved back to beginning of file.

python/turbine_models/tests/stateless_llama_test.py
renxida: done

python/turbine_models/tests/stateless_llama_test.py
renxida: fixed

python/turbine_models/tests/stateless_llama_test.py
renxida: resolved

python/turbine_models/tests/vmfb_comparison.py
renxida: fixd!

renxida:
Requested changes addressed, just need a bigger CI (we currently have 86GB of mem requirement, the CI machine has 62, and a 128 GB ci machine would probably be pretty future-proof)

dan-garvey (approved):

PR #225, opened 2023-12-07
Adds tests for gen_external_params.py quantize function

IanNod:


dan-garvey:
doubles the time of the ci but I dont see a way to avoid that. lgtm

dan-garvey (approved):

IanNod:
> doubles the time of the ci but I dont see a way to avoid that. lgtm

Interesting. Locally took about 1.5 min. Used a smaller opt model for testing to speed it up, maybe I can look to find something even smaller that works that can speed it up eventually

PR #237, opened 2023-12-12
Seperates out runtime

IanNod:
- Adds model_runner.py for setting up iree runtime config to reduce repeated code
 - Creates runner script for llama, clip, unet, and vae to cleanup and simplify the larger scripts


python/turbine_models/custom_models/llama_runner.py
dan-garvey, line 1: Rename to llm or transformer runner?

python/turbine_models/custom_models/sd_inference/vae_runner.py
```diff
@@ -0,0 +1,119 @@
...
+    "--hf_auth_token",
+    type=str,
+    help="The Hugging face auth token, required for some models",
+)
+parser.add_argument(
+    "--device",
+    type=str,
+    default="local-task",
+    help="local-sync, local-task, cuda, vulkan, rocm",
+)
```
dan-garvey, line 39: Instead of copy pasting these we can have a single generic_flags.py and inherit parser from it, adding unique flags in these individual runners.

python/turbine_models/model_builder.py
```diff
@@ -40,11 +40,11 @@ def build_model(self) -> None:
         """
         # TODO: check cloud storage for existing ir
         self.model = self.auto_model.from_pretrained(
-            self.hf_id, use_auth_token=self.hf_auth_token, config=self.auto_config
+            self.hf_id, token=self.hf_auth_token, config=self.auto_config
         )
         if self.auto_tokenizer is not None:
             self.tokenizer = self.auto_tokenizer.from_pretrained(
-                self.hf_id, use_auth_token=self.hf_auth_token
+                self.hf_id, token=self.hf_auth_token
```
dan-garvey, line 47: Nice, glad I'm not the only one who reads warnings lol

python/turbine_models/model_runner.py
```diff
@@ -0,0 +1,30 @@
+import argparse
+import sys
+from iree import runtime as ireert
+
+
+class vmfb_runner:
```
dan-garvey, line 6: nit: CamelCase for classes

dan-garvey (changes requested):

python/turbine_models/custom_models/sd_inference/vae_runner.py
IanNod: Yeah, was thinking of that but looks like flags are kind of in flux with things being added like batch, height, and width to the sd models recently and wanted to let it settle a bit. Will add a todo to address

dan-garvey (approved):

PR #260, opened 2023-12-15
(turbine-models) update diffusers version requirement

monorimet:
@aviator19941 do we have this pinned to such an old version intentionally?

### DIFF
--- a/python/shark_turbine/dynamo/passes.py
+++ b/python/shark_turbine/dynamo/passes.py
@@ -1,6 +1,6 @@
 import torch
 from torch.fx.experimental.proxy_tensor import make_fx
-from torch._decomp import get_decompositions
+from torch._decomp import get_decompositions, register_decomposition
 from torch.func import functionalize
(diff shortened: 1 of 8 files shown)


aviator19941:
> @aviator19941 do we have this pinned to such an old version intentionally?

Yeah, bumping the diffusers version causes unet.py and vae.py to fail because diffusers then introduces `[AT::_SCALED_DOT_PRODUCT_FLASH_ATTENTION](https://pytorch.org/cppdocs/api/function_namespaceat_1a58cab25dce5b7631a1efb0fdd17d2f01.html)` which the tracing fails on for the c10::SymInts.

aviator19941 (approved):

aviator19941:
Sorry I added some cleanup stuff in here too if that's ok with you @monorimet

monorimet:
> Sorry I added some cleanup stuff in here too if that's ok with you @monorimet

On the contrary, much appreciated.

PR #268, opened 2023-12-16
Adding a Resnet 18 Example

zjgarvey:
This pull request adds a simple resnet-18 example with dynamic batches.


examples/resnet-18/resnet-18.py
```diff
@@ -0,0 +1,78 @@
+import numpy as np
```
IanNod, line 1: minor: A README.md would be extremely helpful to add for this

examples/resnet-18/requirements.txt
```diff
@@ -0,0 +1,3 @@
+transformers
```
IanNod, line 1: Maybe worth considering moving resnet from examples to python/turbine_models in some form to enable use outside turbine?

examples/resnet-18/resnet-18.py
```diff
@@ -0,0 +1,78 @@
...
+# this will get passed to the compiled module as a jittable function
+def forward(pixel_values_tensor: torch.Tensor):
+    with torch.no_grad():
+        logits = model.forward(pixel_values_tensor).logits
+    predicted_id = torch.argmax(logits, -1)
+    return predicted_id
+
+
+# A dynamic module for doing inference
+class RN18(CompiledModule):
```
IanNod, line 40: minor: worth mentioning this is an aot example somewhere

examples/resnet-18/resnet-18.py
```diff
@@ -0,0 +1,78 @@
...
+    for num in id:
+        print(model.config.id2label[num])
+
+
+# not sure what the point was of the dynamic dim constraint
+# also amusing that random tensors are always jellyfish
+x = torch.randn(17, 3, 224, 224)
+x[2] = pixel_tensor
+y = shark_infer(x)
+print_labels(y.to_host())
```
IanNod, line 78: Should add a test for this to ensure we don't break our example models

examples/resnet-18/resnet-18.py
zjgarvey: Hi, thanks for the comments. I added a README.md with some instructions.

examples/resnet-18/resnet-18.py
zjgarvey: Okay, I mentioned that in a few places.

examples/resnet-18/resnet-18.py
zjgarvey: I added something like a test, but I'm not sure if it is what you were requesting. I downloaded a bigger dataset and compared the results of the compiled module with the non SHARK-Turbine forward function.

IanNod: I was more thinking unit tests that will run on the CI so anytime this code is updated it will be checked for any kind of failures. An example of the aot mlp example can be found here for reference https://github.com/nod-ai/SHARK-Turbine/blob/main/tests/examples/aot_mlp_test.py

zjgarvey: Amusingly enough, this example broke today upon updating the pip release from 9.2 to 9.3.

IanNod: That is unfortunate, my condolences. This does highlight the fact that the tests are important as whatever change broke this example would have failed and would need to be addressed to get merged.

It may be helpful if you can post an issue of what the error is and maybe a minimal reproducer and we can get more eyes on the problem

zjgarvey: Will do! Thanks for all the comments so far, @IanNod

IanNod: > Will do! Thanks for all the comments so far, @IanNod

Of course, thanks for the work you are doing!

dan-garvey:
@zjgarvey can you mark your test as xfail? we can merge this for now if you do

dan-garvey (approved):

PR #270, opened 2023-12-18
[custom ops] Begin the scaffolding for dispatch of PyTorch custom ops.

stellaraccident:
This makes it possible for us to directly define regular torch ops in terms of generated MLIR. The resulting ops will be specialized and cached per requirements in their definition and will be compiled for any device that Turbine supports when dispatched against tensors on that device.

It is left to a follow-up to also wire this mechanism in on the AOT side so that compiling programs that contain our own custom ops transparently includes them with no further glue. The scaffolding for this is in place, but this patch is big enough without touching AOT.

This allows users to say something like:

```
        @CustomOp.register
        class identity(CustomOp):
            name = "test_identity"
            signature = "(Tensor self) -> Tensor"

            def select(self, ksel: KernelSelection):
                x = ksel.arg_tensor(0)
                ksel.return_tensor(x.t)

            def generate(self, ksel: KernelSelection, kb: KernelBuilder):
                # This just yields the IR value of kernel input as the output.
                # Effectively in eager mode, this is a `return` from the kernel
                # function.
                kb.yield_results(kb.arg_bindings[0])

        t = torch.tensor([[1, 2, 3]], dtype=torch.int32)
        result = identity(t)
        print("CPU result:", result)
        torch.testing.assert_close(result, t)
```

There will be dedicated `CustomOp` subclasses for our various DSLs that can be used for such things (for more sugar'd use than just open coding IR).


python/shark_turbine/runtime/op_reg/base.py
```diff
@@ -0,0 +1,505 @@
...
+    ) -> Callable:
+        """Class decorator for `CustomOp` implementations.
+
+        The decorator will instantiate the class and then replace it with
+        the callable operation that can be used to invoke the kernel.
+
+        Typical usage:
+
+        ```
+        @CustomOp.register
```
dan-garvey, line 82: I didn't realize you could decorate classes. This must be the way pytorch layers are implemented 🤯

python/shark_turbine/runtime/op_reg/base.py
```diff
@@ -0,0 +1,505 @@
...
+        register_impl: bool,
+    ):
+        name = self.name
+        fq_schema = f"{name}{self.signature}"
+        library.define(fq_schema)
+        self.library = library
+        self.op = _get_library_op(library, name)
+
+        # The meta kernel can be provided by the selection machinery and
+        # do not require a tie-in to the kernel generator, which layers
```
dan-garvey, line 120:
```suggestion
        # does not require a tie-in to the kernel generator, which layers
```

python/shark_turbine/runtime/op_reg/eager.py
```diff
@@ -0,0 +1,23 @@
...
+)
+
+
+def eager_dispatch(ksel: KernelSelection):
+    """Main entry-point for handling dispatch of a selected kernel via a generator."""
+    kb = FreeFuncKernelBuilder.create_module(ksel)
+    ksel.op.generate(ksel, kb)
+    print(f"KERNEL IR (cache_key={ksel.spec_key}): {kb.module_op}")
+
+    # TODO: We need a context stack to
```
dan-garvey, line 22: you left me han

tests/runtime/kernel_reg_test.py
```diff
@@ -0,0 +1,47 @@
...
+
+from shark_turbine.runtime.op_reg import *
+
+
+class KernelRegTest(unittest.TestCase):
+    def testSimple(self):
+        @CustomOp.register
+        class identity(CustomOp):
+            name = "test_identity"
+            signature = "(Tensor self) -> Tensor"
```
dan-garvey, line 20: Will users have to write these fancy formatted strings for each implementation?

stellaraccident: Yes, this is just PyTorch syntax for a kernel registration.

python/shark_turbine/runtime/op_reg/base.py
stellaraccident: One of the many magical things, yes. It was horribly broken in Python 2 so was seldom used until the 3.x transition completed.

stellaraccident:
Still working on some finer grained unit tests to complement the integration test.

tests/runtime/op_reg/kernel_reg_test.py
```diff
@@ -0,0 +1,87 @@
...
+class KernelRegTest(unittest.TestCase):
+    def testSimple(self):
+        @CustomOp.register
+        class identity(CustomOp):
+            name = "test_identity"
+            signature = "(Tensor self) -> Tensor"
+
+            def select(self, ksel: KernelSelection):
+                x = ksel.arg_tensor(0)
+                ksel.return_tensor(x.t)
```
raikonenfnu, line 26: Is the general idea to implement the kernels/custom layers in flow_d, linalg_d, arith_d, at the python-mlir level in this `select` fn, and then use ksel.return_tensor at the end to signify the end of the function?

stellaraccident: Not quite. The `select` function is about shape, dtypes, and compile time constants. Basically, each kernel has to look at the incoming signature and cheaply derive:

1. `meta` tensor outputs that describe the expected shape/dtype.
2. Any extra values to be stashed for use by `generate` to actually generate the kernel (optional).
3. A `variant` tag that can further discriminate it in the cache key (optional).
4. Declaring any dimensions desired as static as the default is dynamic (optional).
5. Declare that an argument is a compile time constant and not passed as an IR value (optional).

Based on the metadata that the `select` function determines (either implicitly or explicitly), this will determine a cache key for a specific kernel. If that kernel does not exist in the cache, it will be generated by calling `generate`, compiled, placed in the cache, and used.

Having `select` operate this way is required in order to generate a `meta` kernel impl, which Torch requires for tracing/fx. In addition, although it is not exploited here yet, having it this way is the key to being able to fully pipeline and stream-order allocate sequences of calls without blocking for compilation or invocation. While an advanced feature, that is necessary to compete for eager mode performance, and doing so requires to be able to separate `meta`/selection from compilation and execution.

This raw `CustomOp` class is a low level interface. I'll add some sugar to it to let us do power user things with it (like even just splatting out string-templated ASM and such), but its real power comes when we have, say a `ThreadGridCustomOp` that is completely generated by a chunk of TK vector level DSL (or equiv). And DSL that can do that will be able to auto generate the `select` and `generate` methods and humans using those higher levels won't be thinking about the difference.

stellaraccident:
@dan-garvey Hopefully this made more sense after the f2f. Mind taking another review pass?

python/shark_turbine/runtime/op_reg/eager.py
```diff
@@ -0,0 +1,132 @@
...
+    logger.debug("Kernel invocation %s: %sms", config.key, invoke_time * 1000)
+
+    # Unpack results.
+    results = []
+
+    for i, result_desc in enumerate(ksel.result_descs):
+        meta_tensor_value = result_desc.maybe_tensor_value
+        if meta_tensor_value is None:
+            # Scalar return.
+            raise NotImplementedError("CustomOp scalar return")
```
dan-garvey, line 121: Would this be a good place to contribute?

dan-garvey (approved):
This is great. Are there any examples you would want written to show usage beyond what you're testing here?

python/shark_turbine/runtime/op_reg/eager.py
stellaraccident: Yes - just my style that instead of guessing for things that I don't have an example for, I prefer to fail explicitly. Generally anywhere there is `NotImplementedError` is a case like this.

stellaraccident:
> This is great. Are there any examples you would want written to show usage beyond what you're testing here?

Thanks for the review. Let me get the next patch going where I am going to build out the `iree` op library a bit. Was thinking we could just have PyTorch ops for `trace` and the collective comm ops directly exposed. And then I also need to plumb through to AOT.

tests/runtime/op_reg/kernel_reg_test.py
raikonenfnu: Understood, thanks for the context! :)

PR #287, opened 2023-12-20
add a shark ci

dan-garvey:


dan-garvey:
This only adds support for llms. Will migrate to a shared sh script or python file in shark that runs all desired models once ean adds sd support

monorimet (approved):
LGTM

PR #294, opened 2023-12-21
Vae encode example with test

aviator19941:


python/shark_turbine/dynamo/utils.py
```diff
@@ -0,0 +1,91 @@
...
+        max_k,
+        philox_seed,
+        philox_offset,
+        debug_attn_mask,
+    )
+
+
+# manually add decomposition to bypass the error that comes
+# from VAE encode(inp).latent_dist.sample() failing to symbolically
+# trace from torch fx.
```
stellaraccident, line 65: Can you note which expected Torch stable version will include this by default (basically >current)?

python/shark_turbine/dynamo/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    DeviceLikeType,
+    TensorLikeType,
+)
+import torch._refs as _refs
+from torch._decomp import get_decompositions, register_decomposition
+from torch import Tensor
+from typing import Dict, List, Tuple, Optional
+
+
+@register_decomposition(torch.ops.aten._scaled_dot_product_flash_attention.default)
```
stellaraccident, line 13: I don't love this being here, but I can't think of a better place for now.

stellaraccident (approved):

python/shark_turbine/dynamo/utils.py
aviator19941: noted

PR #296, opened 2023-12-22
Start adding the library of custom ops.

stellaraccident:
Defines a couple of IREE builtins:

* `ops.iree.trace_tensor`
* `ops.iree.trace_tensors`

Extends the infra for better support:

* Adds support for `Tensor[]` arguments to custom ops.


python/shark_turbine/transforms/general/custom_op_expansion.py
```diff
@@ -0,0 +1,248 @@
...
+        results: list[Value],
+        type_converter: NativeTypeConverter,
+    ):
+        super().__init__(op, len(operands))
+        self.operands = operands
+        self.results = results
+        self.type_converter = type_converter
+
+    def arg_tensor(self, arg: int) -> TensorArg:
+        # This is annoying: We have to go from the Torch MLIR type system to the
```
Groverkss, line 128: . Can we not let the custom kernel tell us what the return Torch type will be? I'm not a huge fan of converting from torch-mlir type system.

tests/transforms/general/testdata/custom_op_int_arg.mlir
```diff
@@ -0,0 +1,9 @@
+builtin.module {
+
+func.func @forward() {
+  %i = torch.constant.int 1000
+  torch.operator "torch.expand_custom_op_pass_test.int_arg"(%i) : (!torch.int) -> ()
```
Groverkss, line 5: Should we use File check here?

python/shark_turbine/ops/iree.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2023 Advanced Micro Devices, Inc
```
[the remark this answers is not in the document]
stellaraccident: But this started in 2023 :)

tests/transforms/general/testdata/custom_op_int_arg.mlir
stellaraccident: Good luck with that... We don't have it at this level, and any options to use it are... Not good

python/shark_turbine/transforms/general/custom_op_expansion.py
stellaraccident: I can fiddle with it, but you'll have it anyway by virtue of the eager path through needing to construct meta tensors and all that. If this is something you ever need to see, we can work on it some more. There are already a number of places we need to flip flop like this...

python/shark_turbine/runtime/op_reg/compiler.py
```diff
@@ -87,10 +91,11 @@ def compile_standalone_kernel(
     start = default_timer()
     config = KernelCompileConfig(cache_key, list(device.compile_target_flags))
     kb = FreeFuncKernelBuilder.create_module(ksel, func_name=func_name)
-    ksel.op.generate(ksel, kb)
+    with kb.ip, Location.unknown():
```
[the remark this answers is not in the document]
stellaraccident: I've been putting off writing the stack extraction utility for deriving it. Naive implementations tend to make things worse because they just point to some internal bit of framework detritus and don't interop with error reporting (where you want something like "error happened here but was caused by stuff compiled from here". I know how to do it/have done it before. Just putting off. UnknownLoc is easy to find when it is time to upgrade.

tests/transforms/general/testdata/custom_op_int_arg.mlir
Groverkss: That's sad... Tk could have also used filecheck style tests for fx tracing and codegen.

stellaraccident: We can try again. Last I tried was a few years ago.

python/shark_turbine/ops/iree.py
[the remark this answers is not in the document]
Groverkss: haha, your choice here :)

python/shark_turbine/runtime/op_reg/base.py
```diff
@@ -246,37 +366,116 @@ def arg_int(self, arg: int) -> "IntArg":
...
+    @property
+    def spec_key(self) -> str:
+        """Generates a key that will be the same for all specializations."""
+        return f"int<{self.spec_value}>"
+
+    @property
+    def mlir_type_asm(self) -> str:
+        # TODO: We can have individual kernels constrain this to a narrower
+        # type.
+        return "i64"
```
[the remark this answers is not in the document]
Groverkss: I guess torch-mlir doesnt have an index type

python/shark_turbine/transforms/general/custom_op_expansion.py
Groverkss: I guess if there is no easy solution, let's just do this for now.

tests/transforms/general/testdata/custom_op_int_arg.mlir
Groverkss: there is this python package that duplicates llvm filecheck: https://pypi.org/project/filecheck/ . But I don't know how okay it is to take a dependency like that. In theory, the package shouldn't ever need updates since the syntax is same.

python/shark_turbine/runtime/op_reg/base.py
```diff
@@ -51,21 +57,40 @@
...
     """Users subclass this in order to register a turbine custom op."""
 
     @staticmethod
     def register(
-        op_class: Optional[Type["CustomOp"]],
+        op_class: Optional[Type["CustomOp"]] = None,
         *,
         library: torch.library.Library = TURBINE_LIBRARY,
-        dispatch_key: str = "",
+        dispatch_key: Union[str, Sequence[str], None] = None,
```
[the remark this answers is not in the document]
Groverkss: Oh, i see that this is Union of 3 types, not 2. removing comments.

python/shark_turbine/transforms/general/custom_op_expansion.py
```diff
@@ -0,0 +1,248 @@
...
+        for op in self.ops_to_delete.keys():
+            self.erase_unused_op(op)
+
+    def expand_func(self, func_op: Operation):
+        """Expands custom ops in a traced torch function.
+
+        This finds operations of the form:
+        %0 = torch.operator "torch.ns.op"
+
+        And looks them up in the ALL_CUSTOM_OP_REGS. If it originated from one of those
```
Groverkss, line 71: looks them up in self.reg (ALL_CUSTOM_OP_REGS is just the default value)

python/shark_turbine/transforms/general/custom_op_expansion.py
```diff
@@ -0,0 +1,248 @@
...
+            op,
+            type_converter=self.type_converter,
+            module_body=module_body,
+            symbol_table=self.symbol_table,
+        )
+        with kb.ip, kb.location:
+            op_reg.generate(ksel, kb)
+        assert kb.yielded, "Custom op generation did not yield_results()"
+
+        self.delete_op(op)
```
Groverkss, line 105: So can my custom op not have results? Shouldn't we be replacing all uses here?

Groverkss: ok i see that yield_results does the replacing.

Groverkss (approved):
Overall, the approach makes sense to me. The user interface seems good and the expand pass approach also seems sensible.

tests/runtime/op_reg/kernel_aot_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def forward(self, x: torch.Tensor):
+        x = self.layer0(x)
+        x = torch.sigmoid(x)
+        ops.iree.trace_tensor("LAYER0", x)
+        x = self.layer1(x)
+        x = torch.sigmoid(x)
+        ops.iree.trace_tensor("LAYER1", x)
+        x = self.layer2(x)
+        x = torch.sigmoid(x)
```
raikonenfnu, line 35: This is cool! would we be using these instead of compiledModule in the future?

stellaraccident: Different use cases. But this unifies things so that you can tie arbitrary ir shenanigans to pytorch ops and have them work the same whether in eager dispatch, torch.compile, or aot/export paths

python/shark_turbine/ops/iree.py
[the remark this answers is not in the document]
stellaraccident: Since the whole codebase just has start date, I'll stick with that. Can update all at once if needed.

python/shark_turbine/runtime/op_reg/base.py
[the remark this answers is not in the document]
stellaraccident: Nope. We're going to have to burn down this debt someday...

python/shark_turbine/runtime/op_reg/eager.py
```diff
@@ -50,15 +50,27 @@ def eager_dispatch(ksel: KernelSelection):
...
-        tensor_arg = arg_desc.maybe_tensor_value
-        if tensor_arg is None:
-            continue
-        torch_device = tensor_arg.device
-        device = lookup_device_from_torch(torch_device)
-        if device is not None:
-            break
+        if not arg_desc.is_list:
+            if arg_desc.ir_arity:
+                # One arg has maybe_tensor_value as a single element (common case).
```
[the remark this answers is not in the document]
stellaraccident: I think this was correct but unclear. Changed the condition to check for == 1 for clarity and added a comment to the else branch about this being optional omitted case.

python/shark_turbine/runtime/op_reg/base.py
```diff
@@ -409,10 +642,33 @@ def __init__(
...
+            for d in ksel.result_descs:
+                if not d.is_list:
+                    if d.ir_arity == 1:
+                        result_types.append(IrType.parse(d.mlir_type_asm))
+                    else:
+                        continue
+                else:
+                    # for i in range(arity):
+                    #     result_types.append(IrType.parse(d.mlir_type_asm[i]))
+                    raise AssertionError("NYI: arity > 1 results")
```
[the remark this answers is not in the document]
stellaraccident: Going back to 2023 when the copyright says this was started, I think I started writing this, couldn't come up with a test case and stopped. Will change it to just be NYI.

PR #299, opened 2023-12-23
[WIP][Llama2] Add KVCache for prefill stage + interactive chat mode in llm_runner + StreamingLLM.

raikonenfnu:
Currently our decode initialization/prefill stage is suboptimal compared to the decoding phase. When the token length is rather small, our initalization/prefill is rather quick, but once it gets a bit higher then it starts becoming rather slow. 

Around 2.2-2.8 second at the 500 token mark, and around 5-5.5 seconds when the token len/history is at 1000. This is not very good for multi-dialogue/interactive prompting cases.

KV-Cache has always been known to help our perf during decoding phase. Currently, we are recomputing all the PKV after every new prompt/round/prefill stage, this is rather redundant. This PR introduces the use of KV-Cache at initialization/prefill stage which keeps the time taken to first token/prefill stage at around 0.2 second even at > 1000 token mark.

This PR has been extended to introduce streamingLLM functionality, this will allow us to generate infinite tokens under controlled memory growh. 

Future work for streamingLLM:

- [ ] Make window size configurable through python, everything is there but we'd need to initialize with a default value which would only be possible after we let `_create_initial_value` to take in initial value from GlobalAttribute somewhere [here](https://github.com/nod-ai/SHARK-Turbine/blob/18e8a4100b61adfd9425dd32f780dc5f90017813/python/shark_turbine/aot/support/ir_utils.py#L284-L316) . 
- [ ] Get flow.move to enable overlap of sliding window and src of data. (Currently need to evict when it's at least 2X size of window)
- [ ] Introduce Rerotation of RoPE to as seen [here](https://github.com/huggingface/transformers/blob/c2d283a64a7f33547952e3eb0fa6533fc375bcdd/src/transformers/cache_utils.py#L213-L218) to remove invasive modification of LlamaAttention module for streamingLLM.

This PR also introduce:
1.Set capabilities of GlobalScalars
2.Inheritance of exports/globals for CompiledModule subclasses.
3.READMEs for llm_runner and stateless_llama
4.e2e test refactoring


python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -149,6 +151,33 @@ def run_initialize(self, x=AbstractTensor(BATCH_SIZE, None, dtype=torch.int64)):
                 )
             return token
 
+        def run_cached_initialize(self, x=AbstractTensor(BATCH_SIZE, None, dtype=torch.int64)):
```
raikonenfnu, line 154: Maybe move this inside run_initialize and have it guarded by some if statement?

python/turbine_models/custom_models/llm_app.py
IanNod, line 1: shark-turbine is more for mlir/vmfb generation while shark is the actual frontend user of shark-turbine. I think this would belong more on the shark side and maybe integrate with the cli and webui

stellaraccident (commented):
Cool. Can you add two unit tests for the framework changes?

python/turbine_models/custom_models/stateless_llama.py
```diff
@@ -130,7 +142,7 @@ class StateUpdateModule(CompiledModule):
         else:
             params = export_parameters(mod)
         global_state = export_global(
-            abstractify(global_pkv), uninitialized=True, mutable=True
+            abstractify(global_pkv), uninitialized=False, mutable=True
```
IanNod, line 145: Is this change necessary for this patch? We were initializing this value later and adds a bunch of unnecessary tensor info to the mlir

raikonenfnu: ah good catch, was debugging something with this one.

python/turbine_models/custom_models/llm_app.py
raikonenfnu: That kind of makes sense, but what's the difference between this and the clip_runner, unet_runner, llm_runner?

python/shark_turbine/aot/compiled_module.py
```diff
@@ -343,6 +344,17 @@ def __new__(mcls, name: str, bases, dct, *, export_name: Optional[str] = None):
             if key not in dct:
                 dct[key] = _blackhole_instance_attribute
 
+        # Inheritting methods, globals, and export from parent class.
```
dan-garvey, line 347: nit inheriting*

python/turbine_models/custom_models/llm_app.py
dan-garvey: I suppose the concern is that we already have a lot of this functionality in the shark CLI for llama

dan-garvey: having said that, I don't think its horrible to have something like this here for developer use/testing

dan-garvey (changes requested):
I'm fine with file structure in turbine_models. Please add tests for modify llama and the app

raikonenfnu:
> I'm fine with file structure in turbine_models. Please add tests for modify llama and the app

Hey @dan-garvey I think I am missing something, can you elaborate a bit on how we should test the app?

python/turbine_models/custom_models/llm_app.py
IanNod: Those runners provide functions that shark can make use of and the main function for shark-turbine tests and developers to test with and use. My concern is basically having too much repeated code between shark and shark-turbine that would mostly just cause clutter.

That said I'm not completely against it here, just might want to consider consolidating it on the shark side

IanNod:
> Hey @dan-garvey I think I am missing something, can you elaborate a bit on how we should test the app?

You can look at how we are e2e testing the current stateless_llama app here https://github.com/nod-ai/SHARK-Turbine/blob/main/python/turbine_models/tests/stateless_llama_test.py.

python/turbine_models/custom_models/llm_app.py
raikonenfnu: I see I see, I think SHARK has https://github.com/nod-ai/SHARK/blob/main/apps/shark_studio/api/llm.py, however, I personally prefer the llm_runner.py, because I feel it's a lot more straightforward and easy to look at, which is essentially what llm_app.py is, a llm_runner.py with an interactive prompt. 

Alternatively, any thoughts on me consolidating llm_app.py stuff into llm_runner.py to have an interactive cli mode for development/testing?

python/shark_turbine/aot/compiled_module.py
raikonenfnu: haha nice

python/turbine_models/custom_models/llm_app.py
IanNod: Yeah, if an interactive prompt is helpful for development then I have no problem keeping it. I think consolidating llm_app into llm_runner might be a little cleaner and makes sense.

raikonenfnu: @IanNod I added the chat_mode as part of llm_runner.py, just need to add a `--chat_mode` flag. However, I made sure the original tests were still passing after the refactoring => existing commands and API calls will still work the same. I then refactored and polished up the e2e tests a little bit.

raikonenfnu:
> I'm fine with file structure in turbine_models. Please add tests for modify llama and the app

Thanks Dan! I added the e2e compile and run for streamingLLM vmfb, as well as comparing result between modified_llama and regular llama in torch.

raikonenfnu:
> Cool. Can you add two unit tests for the framework changes?

Thanks for taking a break from your break to look at this Stella, I added the unit tests for the framework changes.

python/turbine_models/custom_models/README.md
IanNod, line 1: Thank you for this! Been on my list for awhile to add a readme

python/turbine_models/custom_models/llm_runner.py
```diff
@@ -46,42 +55,151 @@
 """,
     help="prompt for llm model",
 )
+parser.add_argument(
+    "--chat_mode",
+    action="store_true",
+    help="Runs both turbine vmfb and a torch model to compare results",
```
IanNod, line 61: Nit: Is this the correct help description?

raikonenfnu: good catch!

raikonenfnu: fixed!

IanNod (approved):

dan-garvey (approved):
looks  great! thanks!

PR #307, opened 2024-01-03
Adds an example for resnet50 in half precision.

monorimet:


IanNod (changes requested):
Please add a test for this

PR #350, opened 2024-01-15
[TK] Subgraph Tracing to support control flow

Groverkss:
This PR supports tracing a function within a separate tracer as a "subgraph". This allows us to trace loop bodies instead of unrolling them.

Along with subgraph tracing, this patch adds several features to enable tracing a gemm kernel:

- Instead of doing slicing, add explicit tkl.load/tkl.store ops. Personally, I feel slicing may not be the way to go forward as we always have a constant sized output from the slice. If we go with slicing, we have to analyze if it's constant sized which is not worth it.
- Add support for tkl.constant, tkl.dot, tkl.for_loop.
- Add a new "Vector" class, which is a tensor like class supporting computations over it. I'm not a fan of using pytorch ops directly since I don't get control over the op signature.

I did not add support for eager executing these operations, only compile mode. All of these newly added ops can be eagerly executed, just haven't added the support in this patch.

### DIFF
--- a/python/shark_turbine/kernel/_support/indexing.py
+++ b/python/shark_turbine/kernel/_support/indexing.py
@@ -41,7 +41,7 @@ SubtypeT = TypeVar("SubtypeT")
 class ElementType(ABC):
     @staticmethod
     def cast(something) -> "ElementType":
-        if isinstance(something, torch.dtyp):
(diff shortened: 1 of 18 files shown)


stellaraccident (approved):
This is great! Let's land this and sync up/iterate first thing next week.

PR #361, opened 2024-01-20
Add Fx Pass to remove unbind -> getitem chains

Groverkss:
RNN lowerings provided by pytorch use unbind to convert hidden layer outputs into a list, and then index it later. This doesn't fit well with FxImporter, because it doesn't really have a concept of a list. This pass takes these chains, and converts them into torch.select, so that we don't have lists again.


Groverkss:
I don't think the patch should be landed in its current state. I don't really know where to put users custom fx passes, so I had to plumb them through the jittable pipeline instead. This PR also needs some tests.

PR #375, opened 2024-01-26
[tk] Switch symbolic/shape placeholders to sympy

stellaraccident:
By using sympy instead of the stand-in symbolic support, we get full support for index (partial) evaluation, dynamic dimensions and validation checks.


stellaraccident:
@Groverkss This is passing the tests post my refactor to:

* rebase on sympy
* carry side-car index data along with the graph (so that we can do more analysis/checking at emission)

I dropped the SliceAnalysis thing and inlined it into some more direct cast utilities.

I'm going to keep iterating on this a little bit because now I can add the stuff at the top level to directly make symbols usable in the IR.

python/shark_turbine/kernel/_support/indexing.py
```diff
@@ -569,3 +555,134 @@ def __enter__(self) -> "IndexingContext":
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         context.pop(IndexingContext, self)
+
+
+###############################################################################
+# Symbolic index value type.
+# This type is used in place of `int` to track a symbolic value during
+# traced execution. The type does not carry
```
Groverkss, line 563: Incomplete comment

python/shark_turbine/kernel/_support/indexing.py
```diff
@@ -569,3 +555,134 @@ def __enter__(self) -> "IndexingContext":
...
+
+    def __repr__(self):
+        expr = self.eq_expr
+        if isinstance(expr, IndexSymbol):
+            return f"=={expr}"
+        else:
+            return f"==({expr})"
+
+
+class BoundedRelation(IndexRelation):
```
Groverkss, line 594: I'm not sure if the Index value itself should be tracking this. This seems like something that would come up as an FX analysis pass instead.

python/shark_turbine/kernel/compiler/builder.py
```diff
@@ -124,44 +161,73 @@ def promote_index_to_f32(self, value: Value, to_type: IrType) -> Value:
...
-        return arith_d.subf(lhs, rhs)
-
-    def binary_mod_f32_f32(self, lhs: Value, rhs: Value) -> Value:
-        return arith_d.remf(lhs, rhs)
+ScalarBuilder = _ScalarBuilder()
 
-    def binary_truediv_f32_f32(self, lhs: Value, rhs: Value) -> Value:
-        return arith_d.divf(lhs, rhs)
 
+# class SliceSpecExtractor:
```
Groverkss, line 222: Should we delete these?

python/shark_turbine/kernel/_support/indexing.py
stellaraccident: Yeah, I don't love where this part got to...

Groverkss (approved):
I like this a lot. I think I have some doubts about carrying the range analysis with the value and about slicing, but I think we can fix them later. 

Let's land this and we can iterate on this further.

stellaraccident:
Thanks. I don't like where that was going either. I needed it when I was in the ditch trying to make slice analysis robust... But then I just deleted it and went with something more direct. This revision still needs this for the moment for accessing dynamic grid values, but I don't like that and will scalpel it out.

python/shark_turbine/kernel/_support/indexing.py
stellaraccident: Added a TODO to factor this out.

python/shark_turbine/kernel/compiler/builder.py
stellaraccident: ah, missed that one in my pass through looking for random crap I left lying around.

PR #387, opened 2024-02-01
Add controlnet and controlled unet

gpetters94:
The preprocessors (except canny) need some more work, so they'll go in a separate PR.


python/turbine_models/custom_models/sd_inference/unet.py
```diff
@@ -100,7 +182,7 @@ def export_unet_model(
     if hf_model_name == "stabilityai/stable-diffusion-2-1-base":
         encoder_hidden_states_sizes = (2, 77, 1024)
 
-    sample = (batch_size, unet_model.unet.config.in_channels, height // 8, width // 8)
```
monorimet, line 103: Do we not use the scaling as such for controlled unet? Can this be made conditional?

monorimet (changes requested):
Nice, left a comment -- will also need flag for precision (f16 is only viable path for current GPU optimizations)

monorimet:
> do we not need this anymore?

we definitely need it, though controlnet is lower priority until we have base sd/sdxl inference (+ img2img) locked down

PR #388, opened 2024-02-02
[stateless-llm]Query num_key_value_heads when available to support GQA models.

raikonenfnu:
Models with GQA implemented would not have the same number of heads for K,V vs Query. Hence we need to query `num_key_value_heads` attribute to see if we require different value for KV cache size.

Not all model has `num_key_value_heads` as a part of their config such as QWEN. Phi has it, but it is set to null, hence the code is structured this way to handle those cases.


IanNod (approved):
Looks good to me

PR #389, opened 2024-02-02
Update runners.

raikonenfnu:


stellaraccident:
Looks like the CPU builder doesn't have enough memory

raikonenfnu:
> Looks like the CPU builder doesn't have enough memory

I see I see, I don't have the full list of other runners. Do you have a runner in mind that would be good for this? :)

PR #392, opened 2024-02-04
Fix mlir printer arguments to match new signatures

nirvedhmeshram:
The function signatures seem to have changed after 
https://github.com/llvm/llvm-project/pull/75099

without this change I get
```
TypeError: print(): incompatible function arguments. The following argument types are supported:
    1. (self: iree.compiler._mlir_libs._mlir.ir._OperationBase, state: mlir::python::PyAsmState, file: object = None, binary: bool = False) -> None       
    2. (self: iree.compiler._mlir_libs._mlir.ir._OperationBase, large_elements_limit: Optional[int] = None, enable_debug_info: bool = False, pretty_debug_info: bool = False, print_generic_op_form: bool = False, use_local_scope: bool = False, assume_verified: bool = False, file: object = None, binary: bool = False) -> None
```
I think the file vraible is being confused with the AsmState variable.

### DIFF
--- a/python/shark_turbine/aot/exporter.py
+++ b/python/shark_turbine/aot/exporter.py
@@ -79,7 +79,7 @@ class ExportOutput:
             if file_path.suffix == ".mlirbc":
                 self.mlir_module.write_bytecode(f)
             else:
-                self.mlir_module.print(f, binary=True)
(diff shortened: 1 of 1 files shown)


stellaraccident (approved):
Probably API drift. Better to be explicit like this. Thanks

PR #393, opened 2024-02-05
[tk] Expose dtypes

Groverkss:


stellaraccident (approved):
Nice! I like this. I think the conversion rules as you have them is sensible for this level.

PR #402, opened 2024-02-06
Add a guarded sdpa_cpu torch decomposition and unbind.int

dan-garvey:
Once we migrate default torch version these will be hit by existing tests.

### DIFF
--- a/core/shark_turbine/dynamo/passes.py
+++ b/core/shark_turbine/dynamo/passes.py
@@ -53,6 +53,7 @@ DEFAULT_DECOMPOSITIONS = [
     torch.ops.aten._scaled_dot_product_flash_attention.default,
 ]
 
+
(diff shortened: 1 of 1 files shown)


aviator19941 (commented):
I needed to add torch.ops.aten.unbind.int to the decomposition list when using torch 2.1.0. Were you able to get it to decompose without adding it for 2.1.0?

dan-garvey:
I needed it for torch>2.1 just to pass tests
if it exists in 2.1.0 we can just add it as default

dan-garvey:
Looks like this doesn't work, I'll look at it tomorrow

aviator19941 (approved):

gpetters94 (approved):

dan-garvey:
Not required

PR #409, opened 2024-02-08
[serving] Scaffolding for llm serving.

stellaraccident:
(needs bump for IREE runtime updates)


serving/turbine_serving/llm/testing/fake_v1_module.py
```diff
@@ -0,0 +1,123 @@
...
+                    for i in range(bs):
+                        device_array[i, 0] = i + 1
+
+                    signal_fence.signal()
+                except Exception as e:
+                    signal_fence.fail(str(e))
+
+            threading.Thread(target=run).start()
+
+            result_buffer = device.allocator.allocate_buffer(
```
benvanik, line 83: if you have it exposed you'll want this to be an async alloca (and if not, would be worth getting that exposed)
any alloc you can make async the better - sync deallocate is still fine, but the best is to also get async dealloca so that we can reclaim the memory ASAP
(we need an ABI attribute to let the compiler know when it can async dealloca any inputs to a function, and by default it always returns async alloca'ed results, and when we line that up right it means in the steady state we shouldn't be doing any sync alloc/dealloc)

benvanik (approved):

serving/turbine_serving/llm/testing/fake_v1_module.py
stellaraccident: Thanks - this is just a mock and the real thing pools buffers too. But point taken.

PR #429, opened 2024-02-12
[TK] Add a TestLaunchContext for generating test dispatch IRs

Groverkss:
This patch changes the way we do grid bindings. It is now expected that the IndexingContext will have the workload values and the grid on construction will try to build it's dims using them. Currently, all symbols are assumed to have a constant value, so the grid is able to map workgroup calculations to constants.

Ideally, we would have support to parameterize the kernel by some symbols, which would be passed as workload bindings (We have no support for this currently).


core/shark_turbine/kernel/_support/indexing.py
```diff
@@ -122,20 +115,31 @@ class Grid(metaclass=_GridMeta, symbolic_shape=None):
...
     rank: int
 
-    def __init__(self, *dims: int):
-        rank = len(dims)
-        if self.symbolic_shape is not None:
-            if rank != len(self.symbolic_shape):
-                raise ValueError(
-                    f"Cannot create {type(self)}({', '.join(str(i) for i in dims)}): mismatched symbolic rank"
-                )
+    def __init__(self):
```
Groverkss, line 122: Instead of passing the grid bindings as inputs, I am now making the grid find out the symbol bindings from the IndexingContext, and if it cannot find a binding for the symbol, it should consider it a workload binding. Currently, this promotion to a workload binding is not implemented and every symbol is assumed to have a constant value.

core/shark_turbine/kernel/_support/indexing.py
```diff
@@ -122,20 +115,31 @@ class Grid(metaclass=_GridMeta, symbolic_shape=None):
...
+        else:
+            self.dims = []
 
-        self.dims = dims
         # Shadow the type rank with the actual, which makes it concrete
         # for the generic case.
-        self.rank = rank
+        self.rank = len(self.dims)
+
+    def __class_getitem__(
```
Groverkss, line 137: This is just to fix a type error. Metaclasses should not have __class_getitem__ on them.

core/shark_turbine/kernel/_support/tracing.py
```diff
@@ -404,9 +412,23 @@ def current() -> "LaunchContext":
             return DebugLaunchContext()
 
     def __enter__(self) -> "LaunchContext":
+        # Push an indexing context with the constand bindings for this launch
+        # context in it.
+        # TODO: Is this the most pythonic way to do this? Ask Stella before
```
Groverkss, line 417: Could use a look here. The LaunchContext by itself now also generates an IndexingContext.

core/shark_turbine/kernel/_support/indexing.py
```diff
@@ -122,20 +115,31 @@ class Grid(metaclass=_GridMeta, symbolic_shape=None):
     """Grid with bounding symbolic shape information in the type."""
 
     symbolic_shape: ClassVar[Optional[SymbolicShapeExpr]]
+    # TODO: dims should also allow dynamic dimensions.
+    dims: list[int]
```
stellaraccident, line 119: Not quite seeing why this belongs statically carried on the *type* like this. It also crosses a line that I was trying to hold: the type should not depend on an indexing context.

Fwiw, the symbolic shape can have expressions that are just constants.

stellaraccident: I think what you may be reaching for is an accessor or property of the type which gets a list of dims for the current indexing context (vs pre evaluating that at type construction time).

core/shark_turbine/kernel/_support/tracing.py
```diff
@@ -404,9 +412,23 @@ def current() -> "LaunchContext":
...
+        context.push(IndexingContext, idxc)
+        for s, val in self.constant_bindings.items():
+            idxc.bind_constant(s, val)
+
         return context.push(LaunchContext, self)
 
     def __exit__(self, exc_type, exc_val, exc_tb):
+        # Pop the indexing context created as part of this launch.
+        # TODO: Is this the most pythonic way to do this? Ask Stella before
+        # landing.
```
stellaraccident, line 429: I think I have a helper for this that the others use.

stellaraccident (commented):
First pass: I need to get this review on a bigger screen for a more detailed look.

core/shark_turbine/kernel/_support/indexing.py
Groverkss: I don't think this is being carried on the type. GridMeta is the type which has symbolic_shape, while Grid is an instance which knows how to resolve these shapes unless I'm understanding what you mean wrong.

stellaraccident: My bad - I was looking right past a GH folding bar and missed the second class definition. Was really confused.

core/shark_turbine/kernel/_support/tracing.py
stellaraccident: I think this is fine.

core/shark_turbine/kernel/compiler/host_codegen.py
```diff
@@ -0,0 +1,59 @@
...
+        func_op = func_d.FuncOp("isolated_benchmark", ftype)
+        arg_locs = [
+            (Location.name(b.name) if b.name is not None else Location.unknown())
+            for b in sig.kernel_buffer_input_bindings
+        ]
+        entry_block = func_op.add_entry_block(arg_locs)
+        with InsertionPoint(entry_block):
+            assert isinstance(entry_block, Block)
+            # Create a flow.dispatch op to the kernel
+
```
stellaraccident, line 51: Nit: Remove the blank line or add it above the comment.

stellaraccident (approved):
LG. I think this is just about at the carrying capacity of what we can do without a pass through for more fine grained tests and code organization.

core/shark_turbine/kernel/compiler/host_codegen.py
Groverkss: fixed

PR #473, opened 2024-02-22
Add additional matching logic to MMGroupQuantRewriterPass

Max191:
This adds additional matching logic on `tensor.view` ops required for using MMGroupQuantRewriterPass in the llama.turbine model (used here https://github.com/stellaraccident/llama.turbine/pull/2).

### DIFF
--- a/models/turbine_models/custom_models/stateless_llama.py
+++ b/models/turbine_models/custom_models/stateless_llama.py
@@ -140,11 +140,13 @@ def export_transformer_model(
     )
 
     mapper = {}
+    param_set = set()
     if external_weights is not None:
         if external_weights == "safetensors":
             mod_params = dict(mod.named_parameters())
             for name in mod_params:
                 mapper["params." + name] = name
+                param_set.add(name)
             if external_weight_file:
                 safetensors.torch.save_file(mod_params, external_weight_file)
 
@@ -313,8 +315,12 @@ def export_transformer_model(
     if quantization == "int4" and not compile_to == "linalg":
         from shark_turbine.transforms.quantization import mm_group_quant
 
+        print(param_set)
+        if "lm_head.weight" in param_set:
+            param_set.remove("lm_head.weight")
         mm_group_quant.MMGroupQuantRewriterPass(
-            CompiledModule.get_mlir_module(inst).operation
+            CompiledModule.get_mlir_module(inst).operation,
+            param_names=param_set,
         ).run()
     module_str = str(CompiledModule.get_mlir_module(inst))
     safe_name = hf_model_name.split("/")[-1].strip()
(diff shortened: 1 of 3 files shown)


models/turbine_models/custom_models/stateless_llama.py
```diff
@@ -313,8 +315,12 @@ def evict_kvcache_space(self):
     if quantization == "int4" and not compile_to == "linalg":
         from shark_turbine.transforms.quantization import mm_group_quant
 
+        print(param_set)
```
stellaraccident, line 318: Remove

stellaraccident (commented):
Aside from the test failures, I think I need to see the IR we are trying to match. But this seems like it might be tricky to keep working.

Max191:
> Aside from the test failures, I think I need to see the IR we are trying to match. But this seems like it might be tricky to keep working.

The IR this is matching looks like this:
```
    %_params.blk.0.ffn_gate.weight = util.global.load @_params.blk.0.ffn_gate.weight : tensor<3200x8640xf32>
    %152 = torch_c.from_builtin_tensor %_params.blk.0.ffn_gate.weight : tensor<3200x8640xf32> -> !torch.vtensor<[3200,8640],f32>
    %int8640 = torch.constant.int 8640
    %int3200_171 = torch.constant.int 3200
    %153 = torch.prim.ListConstruct %int8640, %int3200_171 : (!torch.int, !torch.int) -> !torch.list<int>
    %154 = torch.aten.view %152, %153 : !torch.vtensor<[3200,8640],f32>, !torch.list<int> -> !torch.vtensor<[8640,3200],f32>
    %int1_172 = torch.constant.int 1
    %int0_173 = torch.constant.int 0
    %155 = torch.prim.ListConstruct %int1_172, %int0_173 : (!torch.int, !torch.int) -> !torch.list<int>
    %156 = torch.aten.permute %154, %155 : !torch.vtensor<[8640,3200],f32>, !torch.list<int> -> !torch.vtensor<[3200,8640],f32>
    %int3200_174 = torch.constant.int 3200
    %157 = torch.prim.ListConstruct %4, %int3200_174 : (!torch.int, !torch.int) -> !torch.list<int>
    %158 = torch.aten.view %151, %157 : !torch.vtensor<[1,?,3200],f32>, !torch.list<int> -> !torch.vtensor<[?,3200],f32>
    %159 = torch.aten.mm %158, %156 : !torch.vtensor<[?,3200],f32>, !torch.vtensor<[3200,8640],f32> -> !torch.vtensor<[?,8640],f32>
```

This change basically just rewrites the parameter to the viewed type and elides the view (`%154`) as long as the view is just swapping the shape sizes, and the group size evenly divides the new inner shape.

This  pass was already pretty specific to llama, especially since it was matching for a specific parameter name, so I figured that this should be okay for now. Maybe the view rewrite could be moved to its own pass, but this just seemed like the easiest path forward for me, and I like the idea of having fewer passes to call after exporting.

Issue #476, opened 2024-02-24
 Testing Failures with newest transformers version (4.38)


saienduri:
With the latest dependencies (specifically transformers), we are hitting this issue: TypeError: llama_pos_shift_attention_forward() got an unexpected keyword argument cache_position. I manually ran the test models job on the main branch, and it fails as well: https://github.com/nod-ai/SHARK-Turbine/actions/runs/7996420641. Just adding a cache_position default arg doesn't work. Specifically, LlamaRotaryEmbedding definition has changed and use of `seq_len` argument is deprecated and unused in forward. I can pin transformers dependency to a specific version until we figure this out if we want.
```
FAILED models/turbine_models/tests/stateless_llama_test.py::StatelessLlamaChecks::test_rerotated_torch_comparison - TypeError: llama_pos_shift_attention_forward() got an unexpected keyword argument 'cache_position'
FAILED models/turbine_models/tests/stateless_llama_test.py::StatelessLlamaChecks::test_streaming_vmfb_comparison - TypeError: llama_pos_shift_attention_forward() got an unexpected keyword argument 'cache_position'
```

bhbruce:
Hi @saienduri 
Is this problem still happens in newer transformer?

PR #477, opened 2024-02-24
pin transformers for llama testing and mpmath for shark testing

saienduri:
Pin transformers to version 4.37.1 and mpmath to 1.3 to fix the testing failures with the new version until we get it fixed.
Please see #476 for details on reason for testing failure with updated transformers.


stellaraccident (approved):
Before landing, can you please link an issue in the description with your findings. Don't want to lose the diagnosis work you did.

PR #479, opened 2024-02-25
[TK] Refactor ShapedType out of KernelBuffer and Grid

Groverkss:
This patch moves out KernelBuffer and Grid into lang and creates a ShapedType/ShapedDataType class in _support/shaped_type which can be used elsewhere. This is in preparation for VectorType in TK.

This patch also does some other things:
 - Removes the string IndexExpr syntax for KernelBuffer/Grid, which I don't think was very useful.
  - Changes the syntax of KernelBufffer to include data type


Groverkss:
I'm happy to change the syntax of KernelBuffer back. I didn't know before looking at tests that it was possible to use .of

stellaraccident (approved):

PR #480, opened 2024-02-26
[TK] Add test for fused attention and ops needed

Groverkss:
This patch adds a test for f16 attention. The fused_attention kernel needed dtype conversion so that is also added. As a side effect, implicit vector conversion (for unary ops) is now also supported.

### DIFF
--- a/core/shark_turbine/kernel/compiler/builder.py
+++ b/core/shark_turbine/kernel/compiler/builder.py
@@ -32,7 +32,7 @@ from .ir import (
     F64Type,
 )
 
-# TODO: Have a way upstream to check if a floating point type.
+# TODO: Use FloatType from upstream when available.
 FLOAT_BITWIDTHS = {
     "bf16": 16,
     "f16": 16,
@@ -87,6 +87,7 @@ class ModuleBuilder:
 
 class _ScalarBuilder:
     def is_floating_point_type(self, t: IrType) -> bool:
+        # TODO: Use FloatType from upstream when available.
         return str(t) in FLOAT_BITWIDTHS
 
     def is_integer_type(self, t: IrType) -> bool:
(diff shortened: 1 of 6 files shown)


core/shark_turbine/kernel/compiler/builder.py
```diff
@@ -217,10 +238,33 @@ def unary_vector_arithmetic(self, op: str, val: IRProxyValue) -> IRProxyValue:
...
+        # Check bitwidth to determine if we need to extend or narrow
+        from_type = value.type
+        from_bitwidth = self.get_float_bitwidth(from_type)
+        to_bitwidth = self.get_float_bitwidth(to_type)
+        if from_bitwidth < to_bitwidth:
+            return arith_d.extf(to_type, value)
+        elif from_bitwidth > to_bitwidth:
+            return arith_d.truncf(to_type, value)
+        else:
+            raise CodegenError(f"NYI: Cast from {from_type} to {to_type}")
```
Groverkss, line 267: I don't know how to convert between bf16 and fp16

stellaraccident: No one does :) bf16 was made in protest of fp16...

Best that can be done is cast to f32 then to f16. The former is just bit twiddling.

stellaraccident: Or leave it unimplemented with error.

stellaraccident (approved):

core/shark_turbine/kernel/compiler/builder.py
```diff
@@ -87,28 +87,53 @@ def __init__(
 
 class _ScalarBuilder:
     def is_floating_point_type(self, t: IrType) -> bool:
-        return str(t) in FLOAT_TYPES_ASM
+        return str(t) in FLOAT_BITWIDTHS
```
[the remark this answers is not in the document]
Groverkss: oh, cool! Added a TODO, will start using once we bump

core/shark_turbine/kernel/compiler/builder.py
Groverkss: welp, leaving it unimplemented then

PR #482, opened 2024-02-26
Bump IREE to 20240226.813 and adapt to API breaks. 

stellaraccident:
This release pulls in https://github.com/openxla/iree/pull/16486 which makes
substantial changes to torch imports:

* Always generates async code (with a special `$async` suffixed entrypoint that the default entrypoint delegates to).
* Internal structure of the generated code is different, invalidating some tests.

### DIFF
--- a/core/shark_turbine/dynamo/executor.py
+++ b/core/shark_turbine/dynamo/executor.py
@@ -153,7 +153,7 @@ class EagerSpecializedExecutable:
         self,
         user_module: VmModule,
         device_state: DeviceState,
-        entry_name: str = "main",
+        entry_name: str = "main$async",
     ):
         self.user_module = user_module
         self.vm_context = VmContext(
--- a/core/shark_turbine/dynamo/tensor.py
+++ b/core/shark_turbine/dynamo/tensor.py
@@ -53,11 +53,7 @@ from iree.compiler.passmanager import PassManager
 
 from ..importers.fx_importer import FxImporter
 
-DEFAULT_COMPILER_FLAGS = (
-    # Enable asynchronous calling convention.
-    "--iree-execution-model=async-external",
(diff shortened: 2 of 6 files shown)


core/shark_turbine/dynamo/tensor.py
```diff
@@ -53,11 +53,7 @@
 
 from ..importers.fx_importer import FxImporter
 
-DEFAULT_COMPILER_FLAGS = (
-    # Enable asynchronous calling convention.
-    "--iree-execution-model=async-external",
```
raikonenfnu, line 58: Do we not need iree-execution-model=async-external anymore?

core/shark_turbine/dynamo/executor.py
```diff
@@ -153,7 +153,7 @@ def __init__(
         self,
         user_module: VmModule,
         device_state: DeviceState,
-        entry_name: str = "main",
+        entry_name: str = "main$async",
```
raikonenfnu, line 156: Python noob Q: what does the `$` sign do? is this doing something similar to `f"main{async}"` ?

raikonenfnu (commented):
Nice! couple Qs:

core/shark_turbine/dynamo/tensor.py
stellaraccident: No. Async functions are always generated now (with a sync wrapper for default users).

core/shark_turbine/dynamo/executor.py
stellaraccident: Nothing. IREE always exports an "$async" suffixed function when compiling from torch. This is selecting that variant.

stellaraccident:
Discussed offline. The models tests are OOM'ing and seem to be unstable at head.

PR #484, opened 2024-02-27
Add comprehensive tests to test the kernel across available dtypes.

pashu123:
Added softmax and gemm kernel to test across the available float and int dtypes.


core/tests/kernel/coverage_test.py
```diff
@@ -0,0 +1,145 @@
+import torch
+import shark_turbine.kernel as tk
+import shark_turbine.kernel.lang as tkl
+import pytest
+
+
+TKL_TO_TORCH_DTYPE = {
```
Groverkss, line 7: Can we add this to DataType as a method? a from_torch_type and to_torch_type would be good to have.

pashu123: Done. PTAL.

dan-garvey:
is this stale?

pashu123:
> is this stale?

@Groverkss Is this required?

PR #499, opened 2024-03-01
Support for SDXL schedulers + example

monorimet:
so we don't have to carry all these branches


aviator19941 (approved):
should be good thanks!

PR #504, opened 2024-03-02
[TK] Add support to generate tk operations for torch.export

Groverkss:
* Adds `tkl.gen.kernel` decorator to generate a registered/callable kernel.
* The kernel can be used in both AOT and eager contexts.
* Adds some support to kernel registration for uniqueing names like these that are not user-facing and can be defined multiple times.
* Changes the CompiledModule phasing to include a custom op expansion phase. Also defaults it to `IMPORT` instead of `FULL` in a couple of low level places where it still had the old default. This is now consistent with expected, canonical usage but still allows importing to IREE's internal IR if desired (i.e. `FULL`).

### DIFF
--- a/core/shark_turbine/aot/compiled_module.py
+++ b/core/shark_turbine/aot/compiled_module.py
@@ -26,6 +26,7 @@ from ..support.ir_imports import (
     PassManager,
     StringAttr,
 )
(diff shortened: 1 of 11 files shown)


stellaraccident (approved):
Nice - I've pushed some patches with some ergonomic fixes and added a test case with the example you showed me.

PR #505, opened 2024-03-03
[models] Implement KV-cache to have own global for each layer, K, V.

raikonenfnu:
In this implementation, we are splitting up the KV cache to have a global for each layer for each K/V. This will allow for cleaner indexing which will lead to simpler analysis and better perf down the line.

One analysis that we need this for is to check that we have independent reads and writes to KV-cache. We need this to remove redundant copies of KV-cache.  We can also do the same with integrated/big slab but would be much harder since it will require us to analyze dynamic striding and more complex ranges.


raikonenfnu:
FWIW, I ran chat mode on both streaming and non streaming and seem to be running OK on my local system. Seems like just a Runner running out of memory issue again. :/

dan-garvey (approved):
this is cool, makes sense to me.

PR #511, opened 2024-03-05
[export] Add a helper that lets us export multiple entrypoints from an nn.Module

stellaraccident:
I really hope that the `torch.export` folks consider this a real use case to solve for, but for now it seems like it is us and the executorch folks doing gross hacks like this to support export of multiple functions from a module stack.

I went the extra mile on this a bit and went ahead to implement save/load on these stacks, which gives us the ability to do FX tracing once and then iterate on the result to compile. It should be an easier workflow (for dev) considering that Dynamo is so slow for real models.

This is all bleeding edge stuff on Torch, but I need it for LLM so will deal with the sharp edge.


core/shark_turbine/aot/fx_programs.py
```diff
@@ -0,0 +1,325 @@
...
+from pathlib import Path
+from typing import Any, Optional, Union
+
+import functools
+
+import torch
+import torch.nn as nn
+
+# The dynamic_shapes support showed up in the Torch 2.3 timeframe.
+_supports_dynamic_shapes = hasattr(torch.export, "Dim")
```
dan-garvey, line 24: 😉

core/shark_turbine/aot/fx_programs.py
```diff
@@ -0,0 +1,325 @@
...
+            orig_state_dict = dict(ep.state_dict)
+            constants_dict = _get_optional_constants(ep)
+            orig_constants = dict(constants_dict)
+
+            try:
+                # Now unmerge the state_dict and constants by knocking it up against
+                # our running shared state dict.
+                count_deduped += _sharify_state_dict(shared_state_dict, ep.state_dict)
+                count_deduped += _sharify_state_dict(shared_constants, constants_dict)
+
```
dan-garvey, line 100: Have you hit any cases of key collision?

core/shark_turbine/aot/fx_programs.py
```diff
@@ -0,0 +1,325 @@
...
+                # And save our hacked program.
+                save_path = program_files[program_name]
+                torch.export.save(ep, save_path)
```
dan-garvey, line 103: So we are fundamentally saving a dict full of fakeTensors with metadata about what to load from our shared dict on load?

dan-garvey: nevermind, I see it in unsharify, we just check if the loaded value is a SharedStateTensor and redupe

core/tests/aot/fx_programs_test.py
```diff
@@ -0,0 +1,70 @@
...
+    dynamic_shapes = {"x1": {0: batch}, "x2": {0: batch}}
+
+    fxb = FxProgramsBuilder(M())
+
+    @fxb.export_program(args=example_args, dynamic_shapes=dynamic_shapes)
+    def dynamic_batch(module: M, x1, x2):
+        return module.forward(x1, x2)
+
+    @fxb.export_program(args=example_args)
+    def bs1(module: M, x1, x2):
```
dan-garvey, line 52: Does the dynamic shapes annotation here replace the 32s from example_args or does it add a dimension? if the former, 
should `def bs1` be `def bs32` for clarity?

dan-garvey (approved):
Looks useful, even if it just makes it so we can load the traces instead of doing via dynamo.

What are the bleeding edge features this enables?

core/shark_turbine/aot/fx_programs.py
stellaraccident: I *think* it is not possible by construction: the way this is set up has every entrypoint tracing the exact same module path. But that's why I have it asserting. There is another, slightly more complicated way to do this (that the FxImporter does) which would be able to deal with the situation even if the keys were all disjoint. Would just rather not.

stellaraccident:
> Looks useful, even if it just makes it so we can load the traces instead of doing via dynamo.
> 
> What are the bleeding edge features this enables?

For things we currently have, all of the mutability bits descend from ExportedProgram. In terms we could do better with it:

* It internally functionalizes and elminates the need for a bunch of related hackery in jittable.
* Gives us a path for supporting non-persistent buffers.
* Has a way to trace gradients and gradient updates.

PR #513, opened 2024-03-05
Stateless llama testing + SDXL CI Job

saienduri:
This commit cleans up stateless llama testing and helps with memory efficiency by creating the model on the Meta device (with empty weights) and the state dict is then loaded inside it (shard by shard in the case of a sharded checkpoint). The sd test for `vae_encode` throws an error. I took a look at the mlir generated and the `encode_inp func` is there and inputs/return looks valid. Looks like it has to do with iree bump (`FuncConversion` pass). It also adds a CI job to run Jinchen's sdxl script nightly (one failure right now, hopefully fixed soon).

### DIFF
--- a/models/turbine_models/tests/stateless_llama_test.py
+++ b/models/turbine_models/tests/stateless_llama_test.py
@@ -9,7 +9,6 @@ import turbine_models.custom_models.stateless_llama as llama
 import os
 import unittest
 import difflib
-import tracemalloc
 from transformers import AutoTokenizer, AutoModelForCausalLM
 import torch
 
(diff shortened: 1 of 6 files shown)


models/turbine_models/tests/stateless_llama_test.py
```diff
@@ -30,6 +35,22 @@
 Be concise. You are a helpful, respectful and honest assistant. If a question does not make any sense, or is not factually coherent, explain why instead of answering something not correct. If you don't know the answer to a question, please don't share false information. <</SYS>> hi what are you? [/INST]
 """
 
+tokenizer = AutoTokenizer.from_pretrained(
```
stellaraccident, line 38: Note that this does a large download and setup during module loading (i.e. which includes enumerating which tests to run). Such things should be scoped to a setup task).

(I know it was like this so just trying to point out)

stellaraccident (approved):

models/turbine_models/tests/stateless_llama_test.py
saienduri: Ah I see, done

PR #516, opened 2024-03-06
[turbine-llm] Enable dynamic FX tracing of the paged llama model.

stellaraccident:
This was a little rough but got through it. Key changes involve removing data-dependent views.

As per previous, I believe we still have a bug in the per step cache management, and we will need to do a detailed per-step comparison of the cache with a reference.

### DIFF
--- a/llm/turbine_llm/inference/export_paged_llm_v1.py
+++ b/llm/turbine_llm/inference/export_paged_llm_v1.py
@@ -1,0 +1,136 @@
+# Copyright 2024 Advanced Micro Devices, Inc
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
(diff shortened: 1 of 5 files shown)


raikonenfnu (approved):
Overall looks great! I do miss the old IREE-TF style where we still annotate functions, but we do the `FxProgramsBuilder` equivalent on a class/nn.module equivalent. :)

stellaraccident:
> Overall looks great! I do miss the old IREE-TF style where we still annotate functions, but we do the `FxProgramsBuilder` equivalent on a class/nn.module equivalent. :)

We can work more on ergonomics. I'm just racing the clock to be able to do anything with these new APIs, so am going to keep things a bit separated for now... especially since all of this only works on very recent torch nightlies.

PR #546, opened 2024-03-22
Upgrade to PyTorch 2.3.

stellaraccident:
As discussed on Discord, this is a significant upgrade because it is the first stable release that has a fully functional `torch.export.export` with the preferred dynamic shapes support. It is also just prior to nightlies that completely remove support for the old constraints based API, so is therefore a good point to stop for a moment and support both styles. 

This patch makes a number of API changes:

* Issues deprecation warnings if the `constraints=` keyword for jittable is used, otherwise not passing it to PyTorch. This should make jittable not immediately incompatible with later nightlies unless if that feature is used.
* Adds the ability for a `CompiledModule` to directly have an attribute of a `torch.export.ExportedProgram`, allowing the user to pre-export with Torch and then construct a compiled module from that (vs the `jittable` approach where the `CompiledModule` API was directly invoking Torch internals to do so). This defaults to exporting as `public` if given a name not starting with an underscore and private otherwise. Private ExportedPrograms can be called from procedures just as with `jittable`.
* `shark_turbine.aot.export()` now accepts either an `CompiledModule`, `nn.Module`, a or a `torch.export.ExportedProgram`. For the last two, a new `external_params=` bool is available to control whether parameters are inlined or externalized. For an `nn.Module` arguments corresponding to `torch.export.export` are added. Internally, for an `nn.Module`, it simply calls `torch.export.export`. `jittable` is no longer used internally.

Some attempt has been made to be backwards compatible with Torch 2.1.0. New features will not work, but we should be able to support a short buffer window where older pinned systems are not completely broken. The repository prior to this patch will be branched to `torch_2.1`.

Breaking changes:

* ops.iree.trace_tensors (plural) had to be removed because the PyTorch auto functionalization thing has a TODO around lists of tensors. We can add a wrapper that takes a list and invokves trace_tensors multiple times and/or ass a `functional_trace_tensors` which works a bit better with the infra.
* stateless_llama_test.py::test_rerotated_torch_comparison marked as expectedFailure. Filed #560


core/shark_turbine/aot/builtins/jittable.py
```diff
@@ -188,6 +184,17 @@ def resolve_call(
         if self.constraints is not None:
             constraints.extend(self.constraints)
 
+        export_kwargs = {}
+        if len(constraints) > 0:
+            warnings.warn(
+                "Compiling program with the old PyTorch constraints system "
+                "for dynamic shapes is deprecated and will break on PyTorch "
+                "nightlies after the 2.3 release cut (expect either a PyTorch "
+                "warning or excpetion to follow)",
```
dan-garvey, line 193: nit: exception*

dan-garvey (approved):
Some small nits but wow this is amazing. 👏 👏

core/shark_turbine/aot/support/procedural/exported_program.py
```diff
@@ -0,0 +1,276 @@
...
+    ):
+        visibility = self.function_visibility
+        if visibility.value != "private":
+            raise ValueError(
+                f"Currently, only private ExportedPrograms can be called: "
+                f"{self.function_symbol} is {visibility}"
+            )
+
+        # Flatten and convert py args to torch IR values.
+        flat_py_args, args_tree = tree_flatten(((list(py_args),), py_kwargs))
```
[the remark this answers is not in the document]
stellaraccident: Done.

core/shark_turbine/aot/support/procedural/exported_program.py
```diff
@@ -0,0 +1,276 @@
...
+    }
+    for user_output in exported_program.graph_signature.user_outputs:
+        output_node = node_map[user_output]
+        tensor_meta = output_node.meta.get("tensor_meta")
+        fake_val = output_node.meta.get("val")
+        dtype = None
+        if tensor_meta is not None:
+            dtype = tensor_meta.dtype
+        elif fake_val is not None:
+            dtype = fake_val.dtype
```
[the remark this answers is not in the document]
stellaraccident: Yes. Left a TODO to simplify. I retrofitted this from a prior implementation that I know works and didn't want to change it much without more testing to fall back on.

core/shark_turbine/aot/support/procedural/exported_program.py
```diff
@@ -0,0 +1,276 @@
...
+            @property
+            def body(self):
+                return self._op.regions[0].blocks[0]
+
+        return FxImporter(
+            module=FakeModule(module_builder.module_op),
+            config_check=False,
+            py_attr_tracker=module_builder.fx_py_attr_tracker,
+            hooks=hooks,
+        )
```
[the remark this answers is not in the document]
stellaraccident: Yeah, it required a bit of upstream wrangling of https://github.com/llvm/llvm-project/pull/86329 that I didn't want to block on (which is required because of some spidering deps on the Python ExecutionEngine stuff (which is a real mess), which sparse added). Will eventually land that patch/propagate and can drop the fallback.

models/turbine_models/custom_models/README.md
```diff
@@ -39,4 +38,4 @@ python models/turbine_models/custom_models/llm_runner.py --vmfb_path=/path/to/Ll
 2) Interactive CLI chat mode. (just add a --chat_mode flag)
 ```
 python models/turbine_models/custom_models/llm_runner.py --vmfb_path=/path/to/Llama_2_7b_chat_hf.vmfb --external_weight_path=Llama_2_7b_chat_hf_f16_int4.safetensors --device=vulkan hf_auth_token=your_hf_token --chat_mode
-```
\ No newline at end of file
+```
```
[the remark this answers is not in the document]
stellaraccident: Huh, idk. Going to leave it as a minor mystery.

core/shark_turbine/ops/iree.py
```diff
@@ -50,29 +50,15 @@ def _emit_tensor_trace(kb: KernelBuilder, key: str, ts: list[Value]):
...
 @CustomOp.register(library=IREE_LIBRARY)
 class trace_tensor(CustomOp):
-    signature = "trace_tensor(str trace_key, Tensor tensor) -> ()"
+    signature = "trace_tensor(str trace_key, Tensor(a!) tensor) -> ()"
 
     def select(self, ksel: KernelSelection):
         ksel.attr_str(0)
-        ksel.arg_tensor(1)
+        ksel.arg_tensor(1, inplace_tied=True)
+        print("TRACE_TENSOR_SELECT:", ksel)
```
rsuderman: Is it intended to maintain the print statements?

rsuderman (approved):

core/shark_turbine/ops/iree.py
stellaraccident: I whacked them in a followon commit.

core/shark_turbine/ops/iree.py
```diff
@@ -50,29 +50,13 @@ def _emit_tensor_trace(kb: KernelBuilder, key: str, ts: list[Value]):
...
-        ksel.arg_tensor(1)
+        ksel.arg_tensor(1, inplace_tied=True)
 
     def generate(self, ksel: KernelSelection, kb: KernelBuilder):
         key = cast(AttrArg, ksel.arg_descs[0])
         _emit_tensor_trace(kb, cast(str, key.v), [kb.arg_bindings[1]])
-        kb.yield_results()
-
-
-@CustomOp.register(library=IREE_LIBRARY)
```
monorimet, line 65: This trace_tensors op is still in the iree ops test, do we want to converge on trace_tensor and have it handle list targets?

stellaraccident: I'll need to remove it from the tests. Thanks. Unfortunately, new-torch has a TODO on this kind of tensor list on side-effecting ops, so I decided to remove it as the most practical way forward :/

We can have a helper that calls it multiple times (or whatever we want). I was also going to add a `functional_trace_tensor` which returns its value and works a bit better with the infra.

stellaraccident: Noted in the PR description.

PR #548, opened 2024-03-25
update tank and add validation distinction for uploaded names

saienduri:
This commit makes minor changes to the model testing and base models, so that there is a difference between validated and failing mlir names in the turbinetank.

### DIFF
--- a/models/turbine_models/custom_models/sd_inference/utils.py
+++ b/models/turbine_models/custom_models/sd_inference/utils.py
@@ -30,7 +30,7 @@ def largest_error(array1, array2):
 
 
 def compile_to_vmfb(
-    module_str, device, target_triple, max_alloc, safe_name, upload_ir=False
+    module_str, device, target_triple, max_alloc, safe_name
 ):
     flags = [
         "--iree-input-type=torch",
--- a/models/turbine_models/custom_models/sd_inference/utils.py
+++ b/models/turbine_models/custom_models/sd_inference/utils.py
@@ -29,7 +29,9 @@ def largest_error(array1, array2):
     return max_error
 
 
-def compile_to_vmfb(module_str, device, target_triple, max_alloc, safe_name):
+def compile_to_vmfb(
+    module_str, device, target_triple, max_alloc, safe_name, upload_ir=False
+):
     flags = [
         "--iree-input-type=torch",
         "--mlir-print-debuginfo",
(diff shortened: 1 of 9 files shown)


models/turbine_models/turbine_tank/turbine_tank.py
```diff
@@ -44,6 +42,28 @@ def get_short_git_sha() -> str:
...
+    blob_client = blob_service_client.get_blob_client(container_name, old_blob_name)
+    new_blob_client = blob_service_client.get_blob_client(container_name, new_blob_name)
+
+    blob = blob_client.from_connection_string(
+        conn_str=connection_string,
+        container_name=container_name,
+        blob_name=blob_client.blob_name,
+    )
+    if not blob.exists():
+        print("Blob to be renamed does not exist.")
```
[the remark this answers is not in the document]
dan-garvey: you dont have to address the other prints in this file in this patch, but it would be good to replace them all at once

dan-garvey (changes requested):

models/turbine_models/custom_models/sd_inference/utils.py
```diff
@@ -83,7 +85,10 @@ def compile_to_vmfb(module_str, device, target_triple, max_alloc, safe_name):
     with open(f"{safe_name}.vmfb", "wb+") as f:
         f.write(flatbuffer_blob)
     print("Saved to", safe_name + ".vmfb")
-    exit()
+    if upload_ir:
+        return
+    else:
```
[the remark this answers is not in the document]
saienduri: It's the only way I am able to return the `blob_name` at the end of the export. If I stick with the exit(), it doesn't allow the export to return anything. And, it seems that the exit() is required as the `sd_test` is expecting the exit to be raised, so this change allowed me to bypass that without changing any existing behavior.

models/turbine_models/tests/sd_test.py
```diff
@@ -198,6 +217,8 @@ def testExportUnetModel(self):
                 "cpu",
                 upload_ir=UPLOAD_IR,
             )
+            if UPLOAD_IR:
```
dan-garvey, line 220: If we exit here when UPLOAD_IR, how do we ever upload the ir?

models/turbine_models/tests/sd_test.py
```diff
@@ -230,13 +251,17 @@ def testExportUnetModel(self):
         )
         err = utils.largest_error(torch_output, turbine)
         assert err < 9e-5
+        if UPLOAD_IR:
```
dan-garvey, line 254: like how is this code ever reached?

saienduri: If we are uploading IR, we don't exit() and return the blob name. So, we have to manually raise this exit to make the test happy.

dan-garvey (approved):
Thanks for addressing comments!

PR #574, opened 2024-03-28
Enable decompositions for aot.export and torch.export helpers.

stellaraccident:
This was accidentally left out of the PyTorch 2.3 upgrade rework.

Adds APIs to manage decompositions with a context manager so that we don't thread it through every API permutation.

Exposes a bug in PyTorch 2.3 for decompositions with programs that use python registered ops: https://github.com/pytorch/pytorch/issues/122752 (I have included a really gross workaround).

### DIFF
--- a/core/shark_turbine/dynamo/decompositions.py
+++ b/core/shark_turbine/dynamo/decompositions.py
@@ -1,0 +1,128 @@
+# Copyright 2024 Advanced Micro Devices, Inc
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from typing import Callable, Dict, List, Optional, Sequence, Union
+
+import contextlib
+import threading
+
+import torch
+from torch._decomp import get_decompositions, remove_decompositions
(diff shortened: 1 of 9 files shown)


monorimet (approved):
Looks good to me, will fix sdpa decomp on SD branch

core/shark_turbine/dynamo/decompositions.py
```diff
@@ -0,0 +1,129 @@
...
+        aten._log_softmax,
+        aten.nll_loss_forward,
+        aten.nll_loss_backward,
+        aten._to_copy,
+        aten._log_softmax_backward_data,
+        aten.lift_fresh_copy.default,
+        aten._unsafe_index.Tensor,
+        aten.unbind.int,
+        # decompositions added manually in this file
+        aten._scaled_dot_product_flash_attention.default,
```
IanNod, line 119: We want to get rid of this decomp, right? I assume this is a hold over and planned to remove in a future (Ean's SDXL) patch?

IanNod (approved):

core/shark_turbine/dynamo/decompositions.py
stellaraccident: I would assume so. It is just copy-paste in this patch at head.

PR #596, opened 2024-04-03
add verbose flag for models pytest

dan-garvey:


stellaraccident (approved):
Thanks, been meaning to do this but always in the middle of something else.

saienduri (approved):

Issue #125, opened 2023-10-31
[tracking] AOT Export Completion


stellaraccident:
The `shark_turbine.aot.export` API is minimally functional but needs work for completion:

*

PR #597, opened 2024-04-03
Support auto externalization of tensors/params/buffers and normalize export API.

stellaraccident:
This changes the behavior of `aot.export()` by:

* Auto exporting literal tensors that are not otherwise configured. This uses a heuristic to decide whether to inline them or emit them as uniqued, out of line globals.
* Removes the options to `externalize_params=` on `export`. This is now done explicitly by the user prior to calling `export` (see below). This feature was never fully usable like this so it is not believed to have any live uses (people who were using external params features were constructing their own `CompiledModule` exactly as they wanted).
* Adds explicit typing overloads for the four use cases:
  * Export a `torch.nn.Module` (via `torch.export`)
  * Export a pre-constructed `CompiledModule` as a single entry-point
  * Export a pre-exported `torch.export.ExportedProgram` as a single entry-point
  * Export an `FxPrograms` representing a multiple entry-point `torch.export.ExportedProgram` set

Adds APIs:

* `aot.ExternalTensorTrait`: dataclass that contains information about how to externalize the tensor at compile time.
* `CompiledModule.create_from_dict`: Programmatic construction of a `CompiledModule` without using the `class` sugared form.
* `aot.externalize_module_parameters(module)`: Sets `ExportedTensorTrait` on all tensors in the module tree that are parameters or persistent buffers. Such tensors will be emitted to load from an external parameters file when exported.
* `aot.save_module_parameters`: One-stop construction of a parameter archive (IRPA file) from the current values of all parameters and persistent buffers in the module tree.
* `class ParameterArchive`: Wrapper around an IREE `ParameterIndex` with access helpers for interop with torch tensors.
* `class ParameterArchiveBuilder`: Wrapper around an IREE `ParameterIndex` supporting easy construction of IRPA files from torch tensors and modules.

Other behavior changes:

* `{noinline}` removed from global definitions by default (it did default to this).
* turbine_llm misc changes to support auto export of dataset tensors so that they are associated with the originating data file (i.e. gguf file).

Example of exporting a module such that its parameters will be saved to an IRPA file and loaded externally at runtime from it.

```
m = SimpleParamsModule()
externalize_module_parameters(m)
save_module_parameters("/tmp/params.irpa", m)
export(m, args=(torch.empty([128, 20]),)).print_readable()
```

Includes bump of IREE to 20240410.859 (picks up parameters changes to runtime).

### DIFF
--- a/core/shark_turbine/aot/exporter.py
+++ b/core/shark_turbine/aot/exporter.py
@@ -4,10 +4,11 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
+from typing import overload, Any, Dict, List, Optional, Sequence, Tuple, Type, Union
 import io
 from pathlib import Path
 import platform
+import warnings
 
 import torch
 
@@ -25,9 +26,9 @@ from ..support.ir_imports import (
 from .builtins import *
 from .compiled_module import (
     CompiledModule,
-    CompiledModuleMeta,
     ImportPhase,
 )
+from .fx_programs import FxPrograms
 from . import decompositions
 
 __all__ = [
@@ -38,7 +39,12 @@ __all__ = [
 _is_windows = platform.system() == "Windows"
 
 
-ModuleLike = Union[torch.nn.Module, CompiledModuleMeta, torch.export.ExportedProgram]
+ModuleLike = Union[
+    torch.nn.Module,
+    Type[CompiledModule],
+    torch.export.ExportedProgram,
+    FxPrograms,
+]
 SaveableTarget = Union[str, Path, None, Output]
 
 
@@ -150,16 +156,67 @@ class ExportOutput:
             return None
 
 
+@overload
+def export(
+    module: torch.nn.Module,
+    /,
+    *,
+    args: Optional[tuple] = None,
+    kwargs: Optional[Dict[str, Any]] = None,
+    dynamic_shapes: Dict[str, Any] | Tuple[Any] | List[Any] | None = None,
+    module_name: Optional[str] = None,
+    function_name: Optional[str] = None,
+) -> ExportOutput:
+    """Exports a torch.nn.Module.
+
+    This is done by first invoking torch.export.export with args, kwargs,
+    and dynamic_shapes.
+    """
+    ...
+
+
+@overload
+def export(compiled_module: Type[CompiledModule], /) -> ExportOutput:
+    """Exports a CompiledModule and returns the output."""
+    ...
+
+
+@overload
+def export(
+    exported_program: torch.export.ExportedProgram,
+    /,
+    *,
+    module_name: Optional[str] = None,
+    function_name: Optional[str] = None,
+) -> ExportOutput:
+    """Exports a single entry-point module consisting of an ExportedProgram."""
+    ...
+
+
+@overload
+def export(
+    exported_programs: FxPrograms,
+    /,
+    *,
+    module_name: Optional[str] = None,
+) -> ExportOutput:
+    """Exports a multi entry-point ExportedProgram."""
+    ...
+
+
 def export(
     mdl: ModuleLike,
+    /,
     *example_args: torch.Tensor,
     args: Optional[tuple] = None,
     kwargs: Optional[Dict[str, Any]] = None,
     dynamic_shapes: Dict[str, Any] | Tuple[Any] | List[Any] | None = None,
-    external_params: bool = False,
-    global_params: bool = True,
+    module_name: Optional[str] = None,
+    function_name: Optional[str] = None,
 ) -> ExportOutput:
-    """One shot export of an nn.Module or CompiledModule.
+    """Generic export of supported entities.
+
+    See a more specific overload for accepted forms.
 
     This function behaves differently based on the type of the `mdl` argument:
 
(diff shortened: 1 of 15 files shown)


core/shark_turbine/aot/exporter.py
```diff
@@ -178,6 +179,8 @@ def export(
       dynamic_shapes: Dynamic shape specs to pass to torch.export.
       external_params: Whether to declare parameters as external vs inlining
         contents.
+      global_params: Whether to promote module parameters to globals eagerly
```
jpienaar, line 182: Is there perhaps an opt out marker too? (E g., this is either none + opt in or all, wondering if all + opt out option)

stellaraccident: Yeah, this entry point is getting out of control. I'm going to refactor it before landing into a few overloads and better parameterized some of it. There should be fewer options needed if taking the approach of just annotating what you want on prior to export and then having export always so the same thing.

stellaraccident: (i just added this to text the path with no magic, which is what I want to be the default)

stellaraccident: Cleaned all of this up.

core/shark_turbine/aot/support/procedural/exported_program.py
```diff
@@ -235,6 +240,72 @@ def resolve_literal(self, gni: GraphNodeImporter, literal: Any) -> Optional[Valu
...
+        assert isinstance(materialized_global, MaterializedGlobal)
+        return materialized_global
+
+    def _should_lift_tensor_to_global(
+        self, literal: torch.Tensor, external_trait: Optional[ExternalTensorTrait]
+    ) -> bool:
+        if external_trait is not None:
+            return True
+        volume = math.prod(literal.shape)
+        return volume > 1024
```
ScottTodd, line 284: Should this 1024 value be a named constant, or a configurable part of the API?

core/shark_turbine/aot/exporter.py
```diff
@@ -150,15 +156,67 @@ def compile(
...
+def export(
+    module: torch.nn.Module,
+    /,
+    *,
+    args: Optional[tuple] = None,
+    kwargs: Optional[Dict[str, Any]] = None,
+    dynamic_shapes: Dict[str, Any] | Tuple[Any] | List[Any] | None = None,
+    module_name: Optional[str] = None,
+    function_name: Optional[str] = None,
+) -> ExportOutput:
```
ScottTodd, line 169: TIL https://realpython.com/python-asterisk-and-slash-special-parameters/ 

> You use a bare asterisk to define a boundary between arguments that you can pass by either position or keyword from those that you must pass by keyword. You use the slash to define a boundary between those arguments that you must pass by position from those that you can pass by either position or keyword.

> As you know, the asterisk forces you to pass all subsequent arguments by keyword only, while the slash forces you to pass all previous arguments by position. When you use both together, the slash must come before the asterisk.

ScottTodd (approved):

core/shark_turbine/aot/exporter.py
ScottTodd, line 1: I've started looking at updating https://iree.dev/guides/ml-frameworks/pytorch/ to reflect the changes in this PR.

I came across https://github.com/nod-ai/SHARK-Turbine/blob/main/docs/roadmap.md from the https://github.com/nod-ai/SHARK-Turbine README, which linked to https://github.com/nod-ai/SHARK-Turbine/issues/125 `"[tracking] AOT Export Completion"`. Can you make a pass through that tracking issue (and ideally also the other issues linked on the roadmap)?

stellaraccident: Resolved

core/shark_turbine/aot/exporter.py
```diff
@@ -222,23 +277,13 @@ class EpExported(CompiledModule, export_name=mdl.graph_module._get_name()):
...
-            len(example_args) > 0
-            or args is not None
-            or kwargs is not None
-            or dynamic_shapes is not None
-        ):
-            raise ValueError(
-                "If passing a CompiledModule to aot.export, cannot also pass "
-                "args, example_args, kwargs, or dynamic_dims"
-            )
+        assert issubclass(type(mdl), CompiledModule)
```
ScottTodd, line 286: Something is weird here when I try using this. I think the metaclass isn't matching this check. Testing for `aot.compiled_module.CompiledModuleMeta` may be more appropriate?

Could also restructure the branching to
```python
elif isinstance(...):
  # handle CompiledModule
else:
  # print/assert that the type is unsupported
```

repro from https://colab.research.google.com/github/openxla/iree/blob/main/samples/colab/pytorch_aot_advanced.ipynb (which I'm updating to use `iree-turbine` now):

```python
import shark_turbine.aot as aot
import torch

torch.manual_seed(0)

class LinearModule(torch.nn.Module):
  def __init__(self, in_features, out_features):
    super().__init__()
    self.weight = torch.nn.Parameter(torch.randn(in_features, out_features))
    self.bias = torch.nn.Parameter(torch.randn(out_features))

  def forward(self, input):
    return (input @ self.weight) + self.bias

linear_module = LinearModule(4, 3)

example_weight = torch.randn(4, 3)
example_bias = torch.randn(3)

class CompiledLinearModule(aot.CompiledModule):
  params = aot.export_parameters(linear_module, mutable=True)
  compute = aot.jittable(linear_module.forward)

  def main(self, x=aot.AbstractTensor(4)):
    return self.compute(x)

  def get_weight(self):
    return self.params["weight"]

  def set_weight(self, weight=aot.abstractify(example_weight)):
    self.params["weight"] = weight

  def get_bias(self):
    return self.params["bias"]

  def set_bias(self, bias=aot.abstractify(example_bias)):
    self.params["bias"] = bias

export_output = aot.export(CompiledLinearModule)
# assertion error, mdl is not a subclass of CompiledModule
```

Also wondering why unit tests didn't catch this case. I can take a look once I switch from developing in Colab against the release to using the source locally...

ScottTodd: I set up my local environment again. Working on a patch + unit test.

ScottTodd: https://github.com/nod-ai/SHARK-Turbine/pull/615

PR #600, opened 2024-04-08
fix stateless llama testing

saienduri:
This commit fixes the stateless llama testing. Basically, with the setup + teardown approach of pytest unit testing, all the tests were sharing the same model from hugging face. We were running the streaming llama tests before the other tests. In these tests (run_torch_llm and export_transformer_model with streaming_llm=True), we do `enable_llama_pos_shift_attention(model)`, which changes the model we are using. So, this was giving us inaccurate results by the time it came to our base vmfb_comparison test. I created this issue to track and provide more info on the error we are now seeing with torch 2.3 in `test_ test_vmfb_comparison`: https://github.com/nod-ai/SHARK-Turbine/issues/601. Also, marked it as an expected failure for now. (changed runner because we are using previous machine to repro issue for tinygrad folks which can lead to instability and system hangs)

### DIFF
--- a/.github/workflows/test_models.yml
+++ b/.github/workflows/test_models.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         version: [3.11]
-        os: [nodai-amdgpu-w7900-x86-64]
+        os: [nodai-amdgpu-mi210-x86-64]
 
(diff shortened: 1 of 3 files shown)


monorimet (approved):
LGTM!

PR #602, opened 2024-04-09
change for persistent Gentoo runner

saienduri:
Swtiched Gentoo runner from ephemeral to persistent. Now, the python action is not supported for the arch. Minor changes to support the persistent runner.


monorimet (approved):
LGTM! Thanks!

PR #611, opened 2024-04-11
Remove torch pins from SHARK cross-project CI workflow.

monorimet:


dan-garvey (approved):

dan-garvey:
@monorimet please check it looks like a branch is hardcoded somewhere

monorimet:
> @monorimet please check it looks like a branch is hardcoded somewhere

thanks. we fixed the turbine requirement on the SHARK Studio side, so the CI should pass now. Will merge once it does.

monorimet:
It seems that the SHARK-Studio llm api needs to be updated for torch >2.1.0. We should gut the model implementation from studio code as much as possible and use turbine API for actual model import, setup, and execution

PR #615, opened 2024-04-12
Fix assert on `aot.export()` with CompiledModule subclasses.

ScottTodd:
I was hitting an assert trying to pass a subclass of `CompiledModule` through the `aot.export()` API. The code looked correct, but the metaclass code here: https://github.com/nod-ai/SHARK-Turbine/blob/b22dc7f77cab3bc7a7e29ae73398468799edf713/core/shark_turbine/aot/compiled_module.py#L429 could be affecting the check. I looked through https://stackoverflow.com/questions/33347131/determine-if-a-class-in-python-is-a-metaclass for ideas on how to fix.

### DIFF
--- a/core/shark_turbine/aot/exporter.py
+++ b/core/shark_turbine/aot/exporter.py
@@ -26,7 +26,6 @@ from ..support.ir_imports import (
 from .builtins import *
 from .compiled_module import (
     CompiledModule,
-    CompiledModuleMeta,
     ImportPhase,
 )
 from .fx_programs import FxPrograms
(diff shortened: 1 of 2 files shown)


core/shark_turbine/aot/exporter.py
```diff
@@ -282,9 +283,10 @@ def export(
             {(function_name or "main"): exported_program},
             export_name=module_name or "module",
         )
-    else:
-        assert issubclass(type(mdl), CompiledModule)
+    elif issubclass(type(mdl), CompiledModuleMeta):
```
stellaraccident, line 286: Oh I see it. It would have worked originally if `issubclass(MDL, CompiledModule)`. Either way.

core/tests/aot/api_test.py
```diff
@@ -119,6 +119,22 @@ def testTorchExportedProgram(self):
             asm,
         )
 
+    def testCompiledModuleExportedProgram(self):
```
stellaraccident, line 122: Thanks. In my mind, this test existed.

stellaraccident (approved):

core/shark_turbine/aot/exporter.py
ScottTodd: Ah, dropping `type()`. Got it.

PR #622, opened 2024-04-16
Update IREE pip-release-links URL.

ScottTodd:
The `openxla.github.io` link depends on a specific GitHub organization, and the IREE repo will be moving soon (multiple times). The `iree.dev` release links URL is recommended here: https://iree.dev/reference/bindings/python/#prebuilt-packages


monorimet (approved):
LGTM

PR #633, opened 2024-04-18
[llm] Add IRPA transcoding.

stellaraccident:
Primary feature is the addition of `save` and `load` on the `Dataset` class. Previously, it was only possible to create one of these from a GGUF file. Now, we can load from GGUF, save to IRPA, do arbitrary transformations and update, etc.

The rest is the machinery to manage this:

* Layouts types can now be registered and have stable serialization.
* InferenceTensor types can now be registered and have stable serialization.
* Generic PlanarQuantizedTensor which can generically represent the unpacked form of any QuantizedTensor when serializing to IRPA. By default, QuantizedTensors are saved planarized.
* The core ParameterArchive classes were completed to faithfully roundtrip PyTorch tensors.
* There are still things in the CLI that refer to "gguf", but these now operate on any supported archive format (GGUF or IRPA).

Future enhancements include the ability to support incremental IRPA construction with a last-wins/merged approach. This would let arbitrary Theta transformations to be done and appended to new segments of the files. Probably a really useful quantizer feature.

While not done here, this facility makes it pretty trivial to write quantizers and other things since it moves this from bespoke/file-format handling like they are usually based on. Instead, it is just high level transformations on the Theta collection.


llm/turbine_llm/examples/paged_llm_v1.py
KyleHerndon, line 1: Nit: Replace all instances of `gguf_interop` with this change

llm/turbine_llm/tools/dump_gguf.py
```diff
@@ -47,11 +59,7 @@ def main():
             )
         else:
             assert isinstance(tensor, QuantizedTensor), f"Got {type(tensor)}"
-            raw = tensor.raw  # type: ignore
-            print(
-                f"  : QuantizedTensor({tensor.layout_type.__name__})="
-                f"torch.Tensor({list(raw.shape)}, dtype={raw.dtype})"
-            )
+            print(f"  : QuantizedTensor({tensor.layout_type.__name__})=")
```
KyleHerndon, line 62: I think you want to remove the '=' as well

KyleHerndon (approved):

PR #645, opened 2024-04-22
Make new datatypes optional (for compatibility with older torch versions)

stellaraccident:


stellaraccident:
This patch should unbreak HEAD with shark, but I think something is wrong with the checkout on that runner since it is referring to a line that no longer exists.

PR #655, opened 2024-04-25
Remove components that have been upstreamed.

stellaraccident:
* Revises README to explain what is still here.
* Updates CI so that we can use it to run integration tests.


dan-garvey (approved):
skull.jpg

monorimet (approved):
Looks good to me! Separating modelling efforts from the core imports/compiler tooling seems like a wise decision.

PR #656, opened 2024-04-25
Update stream llm to get correct outputs and re-enable rerotated-attention test.

raikonenfnu:
During the update of pytorch/HF, there seem to be a change of how causal mask was being handled. It seems like the attention.forward function used to get a `causal_mask` from the argument as `attention_mask` when is_causal is on. Now it seems like we would need to construct our own mask when `is_causal` is true. This was causing numerical issues in this test as well as on Llama2 qualitatively. 

This PR introduces construction of causal mask, as well as removing unnecessary tensor parallel config checks which simplifies the code quite a bit.


monorimet (approved):
Nice, thanks for this fix! LGTM

PR #706, opened 2024-05-23
Unifies SD pipeline APIs and carries fixes for compile flags, SD1/2

monorimet:


models/turbine_models/custom_models/sd_inference/sd_pipeline.py
```diff
@@ -0,0 +1,748 @@
...
+        # special case for grayscale (single channel) images
+        pil_images = [Image.fromarray(image.squeeze(), mode="L") for image in images]
+    else:
+        pil_images = [Image.fromarray(image) for image in images]
+
+    return pil_images
+
+
+if __name__ == "__main__":
+    from turbine_models.custom_models.sd_inference.sd_cmd_opts import args
```
aviator19941, line 696: Should we use sd_cmd_opts here? The default height and width is 512.

aviator19941 (approved):
There's some commented code in multiple files, did you want to keep those there?

models/turbine_models/custom_models/sd_inference/sd_pipeline.py
aviator19941: I guess it's ok, can just pass in height and width as 1024 through the CL args for SDXL.

monorimet: I kept this as a default since it is the only size supported by all the pipeline's supported models. Since the pipelines are using the same class/API now, they should either use the same args or we need a layer between command line arguments and pipeline API. If we want to do the latter it would be best as a follow-up.

I plan on doing a scrub of things like print statements and commented code as a follow-up -- there are also a few files/functions to deprecate or remove, so that can come as a follow-up to this patch.

models/turbine_models/tests/sdxl_test.py
```diff
@@ -502,98 +400,103 @@ def test04_ExportVaeModelEncode(self):
         np.testing.assert_allclose(torch_output, turbine, rtol, atol)
 
     def test05_t2i_generate_images(self):
```
aviator19941, line 402: seems like this test is failing the CI with this error:
`ValueError: Expected input 4 to be of shape (2, 6) for compiled_unet['run_forward'], got (1, 6).`

PR #729, opened 2024-06-12
Consolidate the stateless llama logic

gpetters-amd:
It's producing vmfbs now, just needs some more cleanup and vmfb runner logic if we want to do that here.

### DIFF
--- a/models/turbine_models/custom_models/llm_cmd_opts.py
+++ b/models/turbine_models/custom_models/llm_cmd_opts.py
@@ -1,0 +1,289 @@
+import argparse
+import os
+from pathlib import Path
+
+
+def path_expand(s):
+    return Path(s).expanduser().resolve()
+
+
+def is_valid_file(arg):
+    if not os.path.exists(arg):
+        return None
+    else:
+        return arg
+
+
+# Note: this is where command-line options for the scripts in this directory
+# are defined along with their defaults. Thus, they should not be referenced
+# within modelling or inference code, only at the entry point to the script.
+
+# We should consider separating out the options that are "model configs" from
+# the options that control the compiler, runtime, and script behavior,
+# when applicable, as the former would best be kept in a separate
+# config or imported from huggingface.
+
+p = argparse.ArgumentParser(
+    description=__doc__, formatter_class=argparse.ArgumentDefaultsHelpFormatter
+)
+
+##############################################################################
+# SDXL Huggingface Options
+##############################################################################
+
+p.add_argument(
+    "--hf_auth_token",
+    type=str,
+    help="The Hugging Face auth token, if required",
+    default=None,
+)
+p.add_argument(
+    "--hf_model_name",
+    type=str,
+    help="HF model name",
+    default="Trelis/Llama-2-7b-chat-hf-function-calling-v2",
+)
+p.add_argument(
+    "--scheduler_id",
+    type=str,
+    help="Scheduler ID",
+    default="Euler",
+)
+
+##############################################################################
+# SDXL Inference Options
+#    These options are used to control runtime parameters for SDXL inference.
+##############################################################################
+
+p.add_argument(
+    "--prompt",
+    type=str,
+    default=" a cat under the snow with blue eyes, covered by snow, cinematic style, medium shot, professional photo, animal",
+    help="Prompt input to stable diffusion.",
+)
+
+p.add_argument(
+    "--negative_prompt",
+    type=str,
+    default="Watermark, blurry, oversaturated, low resolution, pollution",
+    help="Negative prompt input to stable diffusion.",
+)
+
+p.add_argument(
+    "--num_inference_steps", type=int, default=30, help="Number of UNet inference steps"
+)
+
+p.add_argument(
+    "--batch_count",
+    type=int,
+    default=1,
+    help="Number of batches to run for a single prompt",
+)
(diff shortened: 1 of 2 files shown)


dan-garvey:
why are we pulling this to @monorimet's branch? this should be fine standalone, @monorimet can rebase after merging it? That we we get good test coverage?

models/turbine_models/custom_models/stateless_llama.py
```diff
@@ -31,7 +33,7 @@
     "--hf_model_name",
     type=str,
     help="HF model name",
-    default="meta-llama/Llama-2-7b-chat-hf",
+    default="Trelis/Llama-2-7b-chat-hf-function-calling-v2",
```
dan-garvey, line 36: probably not a better default for average caller

models/turbine_models/custom_models/stateless_llama.py
```diff
@@ -489,26 +491,362 @@ def evict_kvcache_space(self):
...
+                ]
+            )
+    flags.extend(llm_model_map[args.hf_model_name]["compile_flags"])
+
+    if not args.pipeline_dir:
+        args.pipeline_dir = "./shark_vmfbs"
+    if not args.external_weights_dir and args.external_weights:
+        args.external_weights_dir = args.pipeline_dir
+
+    sd_pipe = StatelessLlamaPipeline(
```
dan-garvey, line 823: rename

models/turbine_models/custom_models/stateless_llama.py
```diff
@@ -489,26 +491,362 @@ def evict_kvcache_space(self):
...
+        flags,
+        args.attn_spec,
+        args.decomp_attn,
+        args.pipeline_dir,
+        args.external_weights_dir,
+        args.external_weights,
+        args.vae_decomp_attn,
+        args.hf_auth_token,
+    )
+    vmfb, weight = sd_pipe.check_prepared(mlir, vmfb, weight, interactive=False, quantization="int4")
```
dan-garvey, line 843: this should be a flag

models/turbine_models/custom_models/stateless_llama.py
```diff
@@ -489,26 +491,362 @@ def evict_kvcache_space(self):
...
+            input_tensor = input_tensor[:, token_slice:]
+            # if self.streaming_llm and self.model["get_seq_step"]() > 600:
+            if self.model["get_seq_step"]() > 600:
+                print("Evicting cache space!")
+                self.model["evict_kvcache_space"]()
+            token_len = input_tensor.shape[-1]
+            device_inputs = [
+                ireert.asdevicearray(self.device, input_tensor)
+            ]
+            if self.first_input: # or not self.streaming_llm:
```
dan-garvey, line 746: streaming llm commented code? is this from the original?

dan-garvey (changes requested):

models/turbine_models/custom_models/stateless_llama.py
monorimet: might make more sense to rename it to "prepare_pipeline" since this checks for vmfbs, and exports them if they aren't found.

models/turbine_models/custom_models/stateless_llama.py
gpetters-amd: I redid some of the logic to remove non-streaming since I thought our plan was to only support streaming, but I think that's not actually the case. I'll add the support back in.

models/turbine_models/custom_models/llm_cmd_opts.py
```diff
@@ -0,0 +1,187 @@
...
+    default="meta-llama/Llama-2-7b-chat-hf",
+)
+
+##############################################################################
+# Runtime Options
+#    These options are used to control runtime parameters.
+##############################################################################
+
+p.add_argument(
+    "--seed", type=float, default=0, help="Seed for random number/latents generation."
```
IanNod, line 52: llama doesn't need a seed does it?

models/turbine_models/custom_models/llm_cmd_opts.py
```diff
@@ -0,0 +1,187 @@
...
+)
+
+p.add_argument(
+    "--pipeline_dir",
+    type=str,
+    default="",
+    help="Path to location of vmfb files.",
+)
+
+p.add_argument(
```
IanNod, line 62: unnecessary weight flags for llama. We are only using 1 external weight file so could remove external_weights_dir, and I don't think we need external_weight_file below,

models/turbine_models/custom_models/resnet_18.py
```diff
@@ -8,7 +8,7 @@
 from iree.compiler.ir import Context
```
IanNod, line 8: Let keep the separate model updates in separate patches. Makes it easier to track and revert patches if ever needed

models/turbine_models/tests/stateless_llama_test.py
```diff
@@ -90,20 +90,42 @@ def test_vmfb_comparison(self):
 
         upload_ir_var = os.environ.get("TURBINE_TANK_ACTION", "not_upload")
 
-        blob_name = llama.export_transformer_model(
+        # blob_name = llama.export_transformer_model(
```
IanNod, line 93: commented code?

PR #740, opened 2024-06-14
Add ability to include a plugin when creating driver

daveliddell:
`vmfbRunner` now takes an optional `extra_plugin` argument to load an executable plugin while the driver is getting created.

This option might be used, for example, when loading a vmfb that has an external dependency on a native shared library.

The implementation of this new feature takes advantage of the pre-existing `iree.runtime.flags` feature and a new IREE python API function.  Normally, drivers are managed in a cache.  However, setting a flag to specify the plugin has no effect on existing drivers.  The API now has a function for creating a driver independent of the cache, to guarantee that any flags are sure to take effect.

This PR also includes a fix for the problem of the CI using old cached wheels for iree, as recommended by @monorimet.


monorimet (approved):
Nice! Thanks for the push on this. Love that it's just a few lines change on the python side.

PR #766, opened 2024-07-09
sharktank UNet model support.

monorimet:


models/turbine_models/custom_models/sd_inference/vae.py
```diff
@@ -54,97 +55,175 @@ def __init__(
...
+            torch.ops.aten._scaled_dot_product_flash_attention.default,
+            torch.ops.aten.scaled_dot_product_attention,
+        ]
+    with decompositions.extend_aot_decompositions(
+        from_current=True,
+        add_ops=decomp_list,
+    ):
+        fxb = FxProgramsBuilder(vae_model)
+
+        # @fxb.export_program(args=(encode_args,))
```
aviator19941, line 186: remove commented code

models/turbine_models/custom_models/sdxl_inference/unet.py
```diff
@@ -70,9 +72,101 @@ def forward(
...
+        output_path = external_weight_path.split("unet")[0] + "punet_dataset_i8.irpa"
+        ds = get_punet_i8_dataset(
+            results["config.json"],
+            results["quant_params.json"],
+            results["params.safetensors"],
+            output_path,
+            base_params=None,
+        )
+    else:
+        ds = None  # get_punet_dataset(results["config.json"], results["params.safetensors"], base_params=None)
```
aviator19941, line 110: Why is this commented out?

models/turbine_models/custom_models/sdxl_inference/unet.py
```diff
@@ -161,36 +296,63 @@ def export_unet_model(
...
         add_ops=decomp_list,
     ):
-        fxb = FxProgramsBuilder(unet_model)
+        if external_weights:
+            externalize_module_parameters(unet_model)
+        if use_punet:
+            output = export(
+                unet_model,
+                kwargs=example_forward_args_dict,
+                module_name="compiled_unet",
```
aviator19941, line 305: can this be compiled_punet or no?

models/turbine_models/custom_models/sdxl_inference/unet.py
```diff
@@ -70,9 +72,101 @@ def forward(
...
+
+def get_punet_i8_dataset(
+    config_json_path,
+    quant_params_path,
+    params_path,
+    output_path="./punet_dataset_i8.irpa",
+    quant_params_struct=None,
+    base_params=None,
+):
+    from sharktank.models.punet.tools.import_brevitas_dataset import (
```
stellaraccident, line 125: Please don't do it this way. Just import the `main` function and pass it CL args vs using the private internals.

models/turbine_models/custom_models/sdxl_inference/unet.py
```diff
@@ -70,9 +72,101 @@ def forward(
...
+
+    # Apply updates into a new Theta.
+    theta = base_theta if base_theta is not None else quant_theta
+    flat_tensors = theta.flatten()
+    flat_tensors.update(updated_tensors)
+    ds.root_theta = Theta(flat_tensors)
+
+    # TODO: Post-process to introduce fused cross-layer connections.
+
+    ds.save(output_path, io_report_callback=print)
```
stellaraccident, line 164: There is a bug in the Dataset handling code where if you construct a Dataset on the fly and then use that without loading it from disk, it does not annotate the imported tensors, resulting in them being inlined vs referenced. I need to fix that bug, but the workflow in this case shouldn't be doing this anyway.

If you follow the advice in the comment above and use the main() entrypoint to generate the IRPA on disk, then do Dataset.load(irpa_path), that will work.

models/turbine_models/custom_models/sdxl_inference/unet.py
monorimet: It can.

models/turbine_models/custom_models/sdxl_inference/unet.py
monorimet: Haven't implemented it yet. just ran i8 straight through and left some spots for the other configurations. Will replace with TODOs or just finish it after higher priority items

monorimet:
needs https://github.com/iree-org/iree-turbine/pull/40 to land for some metadata pass fixes on clip.

models/turbine_models/custom_models/sd_inference/tokenization.py
```diff
@@ -4,415 +4,171 @@
...
-    prompt_tokens, prompt_weights = pad_tokens_and_weights(
-        prompt_tokens,
-        prompt_weights,
-        max_length,
-        bos,
-        eos,
-        no_boseos_middle=no_boseos_middle,
-        chunk_length=pipe.model_max_length,
-    )
+            print("The following text was removed due to truncation:", removed_text)
```
stellaraccident, line 89: Did you mean to use `print` like this? Generally not great to spew to stdout from a library. Maybe use `warnings.warn(...)` instead?

models/turbine_models/custom_models/sd_inference/unet.py
```diff
@@ -114,76 +113,96 @@ def export_unet_model(
...
-            ),
-            guidance_scale=AbstractTensor(1, dtype=dtype),
+    example_forward_args = [
+        torch.empty(sample, dtype=dtype),
+        torch.empty(1, dtype=dtype),
+        torch.empty(encoder_hidden_states_sizes, dtype=dtype),
+        torch.empty(1, dtype=dtype),
+    ]
+    decomp_list = []
+    if decomp_attn == True:
```
stellaraccident, line 133: Why not just `if decomp_attn:`?

models/turbine_models/custom_models/sd_inference/utils.py
```diff
@@ -331,7 +340,8 @@ def get_wmma_spec_path(target_chip, save_dir, masked_attention=False):
     else:
         return None
     attn_spec = urlopen(url).read().decode("utf-8")
```
stellaraccident, line 342: I suppose this isn't the only place where we are fetching network resources, but wish we weren't doing this... some day.

models/turbine_models/custom_models/sdxl_inference/unet.py
```diff
@@ -67,12 +74,85 @@ def forward(
...
+        Unet2DConditionModel as sharktank_unet2d,
+        ClassifierFreeGuidanceUnetModel as sharktank_CFGPunetModel,
+    )
+    from sharktank.utils import cli
+
+    if precision == "i8":
+        repo_id = "amd-shark/sdxl-quant-models"
+        subfolder = "unet/int8"
+        revision = "82e06d6ea22ac78102a9aded69e8ddfb9fa4ae37"
+    elif precision in ["fp16", "fp32"]:
```
stellaraccident, line 96: FYI - the sdxl-quant-models do have an fp16 variant IRPA file. I use that for various comparisons, etc.

https://huggingface.co/amd-shark/sdxl-quant-models/tree/main/unet/fp16/export

stellaraccident: I guess we can decide whether we want to do the full model download/quantization here or just fetch the IRPA file from the export subdir. If we did the latter, you could drop get_punet_dataset() that is actually building it and just use the one we publish. If doing that, you can eliminate this branch and just change the subfolder and filename or something.

Fine to check in like this since it is working, but can simplify by leaning more on published artifacts.

Now that I'm thinking of it, though, we're going to need a procedure for ML-Perf to rely completely on generated artifacts. Currently, that is being done by providing scripts to regenerate, but then we probably also need to support flags here and elsewhere to side-load.

In any case, let's land and discuss/iterate.

models/turbine_models/custom_models/sdxl_inference/unet.py
```diff
@@ -67,12 +74,85 @@ def forward(
...
+            repo_id=repo_id, subfolder=subfolder, filename=filename, revision=revision
+        )
+
+    results = {
+        "config.json": download("config.json"),
+        "params.safetensors": download("params.safetensors"),
+    }
+    if precision == "i8":
+        results["quant_params.json"] = download("quant_params.json")
+        output_path = external_weight_path.split("unet")[0] + "punet_dataset_i8.irpa"
```
stellaraccident, line 112: What would happen if this were in the directory `C:\Users\unet\Downloads`? Would be better to use pathlib for manipulation of just the stem or something.

models/turbine_models/custom_models/sdxl_inference/unet.py
```diff
@@ -67,12 +74,85 @@ def forward(
...
+            results["params.safetensors"],
+            output_path,
+            results["quant_params.json"],
+            base_params=None,
+        )
+    else:
+        ds = get_punet_dataset(
+            results["config.json"],
+            results["params.safetensors"],
+            output_path,
```
stellaraccident, line 124: output_path looks uninitialized here

models/turbine_models/custom_models/sdxl_inference/unet.py
```diff
@@ -67,12 +74,85 @@ def forward(
...
+
+    cond_unet = sharktank_unet2d.from_dataset(ds)
+    mdl = sharktank_CFGPunetModel(cond_unet)
+    return mdl
+
+
+def get_punet_dataset(
+    config_json_path,
+    params_path,
+    output_path="./punet_dataset_i8.irpa",
```
stellaraccident, line 136: Remove the default. Should always be provided, I think.

models/turbine_models/custom_models/sdxl_inference/unet.py
```diff
@@ -67,12 +74,85 @@ def forward(
...
+    quant_params_path=None,
+    quant_params_struct=None,
+    base_params=None,
```
stellaraccident, line 139: Seems not used/remove (this was a hack/prototype thing that didn't work out anyway).

models/turbine_models/custom_models/sdxl_inference/unet.py
```diff
@@ -67,12 +74,85 @@ def forward(
...
+    mdl = sharktank_CFGPunetModel(cond_unet)
+    return mdl
+
+
+def get_punet_dataset(
+    config_json_path,
+    params_path,
+    output_path="./punet_dataset_i8.irpa",
+    quant_params_path=None,
+    quant_params_struct=None,
```
stellaraccident, line 138: Not used/remove?

models/turbine_models/custom_models/sdxl_inference/unet.py
```diff
@@ -67,12 +74,85 @@ def forward(
...
+    base_params=None,
+):
+    from sharktank.models.punet.tools import import_brevitas_dataset
+
+    import_brevitas_dataset.main(
+        [
+            f"--config-json={config_json_path}",
+            f"--params={params_path}",
+            f"--quant-params={quant_params_path}",
```
stellaraccident, line 147: I don't think this will work for the fp16 version like this (the literal value "None" will try to load a file named "None").

There is an importer right next to this for a stock huggingface dataset which was intended to be used for this case. Should probably switch here, but judging from above, I'm guessing that fp16 punet has not been tested yet. Can flush when it is.

models/turbine_models/custom_models/sdxl_inference/unet.py
```diff
@@ -84,18 +164,29 @@ def export_unet_model(
...
+    if use_punet:
+        submodel_name = "punet"
+    else:
+        submodel_name = "unet"
+    if (not decomp_attn) and use_punet:
+        attn_spec = "punet"
+    elif (not decomp_attn) and "gfx9" in target:
+        attn_spec = "mfma"
+    elif (not decomp_attn) and "gfx11" in target:
+        attn_spec = "wmma"
```
stellaraccident, line 186: This kind of thing makes me sad. Nothing to do about it now.

stellaraccident (approved):
Left a number of comments that would be good to address here or in a followup.

PR #770, opened 2024-07-11
Minor fixes for SDXL/unet

saienduri:
The reason for updating the revision hash is this PR by Stella in sharktank: https://github.com/nod-ai/sharktank/pull/93. Because we are using sharktank TOM, we need to update here too so that it gives sharktank the expected quant_params.json.


monorimet (approved):
LGTM

PR #774, opened 2024-07-12
(SD) Bump punet revision to d30d6ff and enable punet test.

monorimet:
The SHA may need to be updated once the fused qkv model is merged in the amd-shark/sdxl-quant-models repo.


models/turbine_models/custom_models/sdxl_inference/unet.py
```diff
@@ -92,11 +92,11 @@ def get_punet_model(hf_model_name, external_weight_path, precision="i8"):
     if precision == "i8":
         repo_id = "amd-shark/sdxl-quant-models"
         subfolder = "unet/int8"
-        revision = "942e771bf0c2657a8b33380103d04747a75dfa4a"
+        revision = "d30d6ff79abb584bf2addc7866738df5242f315a"
     elif precision in ["fp16", "fp32"]:
         repo_id = hf_model_name
         subfolder = "unet"
-        revision = "76d28af79639c28a79fa5c6c6468febd3490a37e"
+        revision = "d30d6ff79abb584bf2addc7866738df5242f315a"
```
stellaraccident, line 99: Shouldn't need to be updated. I'll just merge to main but the commit should stay the same. 

We can use tags on hf and may want to do that at some point.

stellaraccident (approved):

PR #783, opened 2024-07-21
Integrate int8 tk kernels

nithinsubbiah:


models/turbine_models/custom_models/sdxl_inference/sdxl_cmd_opts.py
```diff
@@ -369,5 +369,11 @@ def is_valid_file(arg):
     help="extra iree-compile options to send for compiling unet. Only use this for testing bleeding edge flags! Any default options should be added to sd_inference/utils.py",
 )
 
+p.add_argument(
```
saienduri, line 372: Can we change this to:

```
p.add_argument(
    "--add_tk_kernels",
    default=False,
    action="store_true",
    help="Flag to add compiled tk kernels.",
)
```

User can just use the flag and not have to explicitly set it True. This is what we have been doing for the bool flags

saienduri (commented):
Nice, I have tested that this tk integration is working and looks good to me
Just needs formatting (`pip install black && black models`)

Next step is to enable this flag and run it in the harness to see if it helps in any way

PR #785, opened 2024-07-22
Changes so that the default run without quant docker will work as well with all the new changes.

saienduri:
This PR makes the necessary changes so that you don't have to run the quantization docker to run the normal harness. With stricter checks for the quant paths and adding a special switch to the save_external_weights so that it is compliant with this change that we need for the docker generated weights https://github.com/iree-org/iree-turbine/commit/cd916ec7aec3e9fcc91898a3b6b93e575148e3f3.


monorimet (approved):
LGTM.

PR #786, opened 2024-07-22
Sync flags to sdxl-scripts repo

saienduri:
Sync flags to sdxl-scripts repo


monorimet (approved):
This isn't really the direction we should be continuing in for managing these things, but for this week let's make it work and think about how to avoid all this yucky python spaghetti for compile flags.

I believe we spoke at some point about having a versioned repository of flagfiles to carry these things -- even if it ends up being "more redundant" to have one-shot configs I'd rather that than use model names in the flags generator.

Putting a bookmark in this so we can revisit after this week.

monorimet:
filed https://github.com/nod-ai/SHARK-Turbine/issues/787.

The CI failure is related to another patch, should be an easy fix to avoid running `os.path.exists()` on `NoneType`

saienduri:
Yeah I fully agree. Felt like the configuration process here was getting too messy while updating

PR #789, opened 2024-07-24
Integration of tk kernels into pipeline

saienduri:
Currently using a link, but Nithin will be pushing the fix to use a file name asap


monorimet (approved):
LGTM

PR #791, opened 2024-07-25
[tk kernel] Add support to match kernel with number of arguments and update kernel links

nithinsubbiah:


saienduri (approved):
LGTM, just formatting

PR #793, opened 2024-07-25
Remove download links for tk kernels and instead specify kernel directory as an argument

nithinsubbiah:


saienduri (approved):
LGTM thanks

PR #798, opened 2024-07-29
Fix the compiled pipeline compilation issue

gpetters-amd:
It compiles the vmfb now, but there are some issues in the script preventing it from running properly. I can fix that in this PR as well or make a new one after this one is merged.

### DIFF
--- a/models/turbine_models/custom_models/sd_inference/sd_pipeline.py
+++ b/models/turbine_models/custom_models/sd_inference/sd_pipeline.py
@@ -727,10 +727,7 @@ class SharkSDPipeline(TurbinePipelineBase):
         for i in range(batch_count):
             if self.compiled_pipeline:
                 image = produce_images_compiled(
-                    samples[i],
-                    prompt_embeds,
-                    negative_embeds,
-                    guidance_scale
+                    samples[i], prompt_embeds, negative_embeds, guidance_scale
                 )
             else:
                 produce_latents_input = [
--- a/models/turbine_models/custom_models/sd_inference/utils.py
+++ b/models/turbine_models/custom_models/sd_inference/utils.py
@@ -36,13 +36,13 @@ MI_flags = {
     ],
     "unet": [
         "--iree-flow-enable-aggressive-fusion",
-        "--iree-flow-enable-fuse-horizontal-contractions=true",
+        "--iree-global-opt-enable-fuse-horizontal-contractions=true",
         "--iree-opt-aggressively-propagate-transposes=true",
         "--iree-codegen-llvmgpu-use-vector-distribution=true",
     ],
     "clip": [
         "--iree-flow-enable-aggressive-fusion",
-        "--iree-flow-enable-fuse-horizontal-contractions=true",
+        "--iree-global-opt-enable-fuse-horizontal-contractions=true",
         "--iree-opt-aggressively-propagate-transposes=true",
     ],
     "vae": [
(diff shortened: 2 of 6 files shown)


models/turbine_models/custom_models/sd_inference/sd_pipeline.py
```diff
@@ -465,13 +465,10 @@ def encode_prompts_sdxl(self, prompt, negative_prompt):
             text_input_ids_list += text_inputs.input_ids.unsqueeze(0)
             uncond_input_ids_list += uncond_input.input_ids.unsqueeze(0)
 
-        if self.compiled_pipeline:
```
IanNod, line 468: Is there any particular reason you are removing the non compiled_pipeline option here?

models/turbine_models/custom_models/sd_inference/sd_pipeline.py
```diff
@@ -624,9 +622,11 @@ def produce_images_compiled(
             sample,
             prompt_embeds,
             text_embeds,
-            guidance_scale,
+            torch.as_tensor([guidance_scale], dtype=sample.dtype),
```
IanNod, line 625: Is this necessary?

models/turbine_models/custom_models/sd_inference/sd_pipeline.py
```diff
@@ -624,9 +622,11 @@ def produce_images_compiled(
...
         ]
-        image = self.compiled_pipeline("produce_img_latents", pipe_inputs)
+        # image = self.compiled_pipeline("produce_img_latents", pipe_inputs)
```
IanNod, line 627: remove commented code

models/turbine_models/custom_models/sd_inference/utils.py
```diff
@@ -36,13 +36,13 @@
     ],
     "unet": [
         "--iree-flow-enable-aggressive-fusion",
-        "--iree-flow-enable-fuse-horizontal-contractions=true",
+        "--iree-global-opt-enable-fuse-horizontal-contractions=true",
```
IanNod, line 39: Is change needed for current IREE?

IanNod (commented):
Can you work on getting this merged into main instead of a separate compiled-pipeline branch? Would need to allow for flag controlling whether to use compile-pipeline or not

models/turbine_models/custom_models/sd_inference/utils.py
gpetters-amd: It seems to be, at least as of the 7/24 nightly. We might be using a newer version in the CI.

models/turbine_models/custom_models/sd_inference/sd_pipeline.py
gpetters-amd: Yep, the compiled pipeline expects an `fp16` tensor while passing this produces an `fp64` one.

models/turbine_models/custom_models/sd_inference/sd_pipeline.py
gpetters-amd: It's removing the compiled_pipeline version - I think this is a relic of previous iterations when the text encoder was going to be part of the pipeline, but since it isn't we can run the same code for compiled and python pipelines.

models/turbine_models/custom_models/sd_inference/utils.py
IanNod: 7/24 is very old. Should test with latest or at least what iree-turbine pulls (which of right now is iree-compiler==20240828.999). Looks like the CI failures you are seeing are also due to flag changes with newer iree versions

PR #799, opened 2024-07-30
Upstreaming MLPerf punet changes, server/harness support.

monorimet:


models/turbine_models/custom_models/sd3_inference/diffusers_ref.py
```diff
@@ -0,0 +1,49 @@
+from diffusers import StableDiffusion3Pipeline
```
IanNod, line 1: Should we integrate this with a test to output the image numerics we can compare against? I know you saw some significant different numerics between cpu and different gpu backends where this may be difficult to directly compare, maybe some FID/CLIP scores?

models/turbine_models/custom_models/sd3_inference/sd3_mmdit.py
```diff
@@ -249,7 +253,7 @@ class CompiledMmdit(CompiledModule):
             hidden_states_shape,
             encoder_hidden_states_shape,
             pooled_projections_shape,
-            init_batch_dim,
+            (1,),
```
IanNod, line 256: is mmdit not working batched?

models/turbine_models/custom_models/sd_inference/sd_pipeline.py
```diff
@@ -190,6 +195,7 @@ def get_sd_model_map(hf_model_name):
         "stabilityai/sdxl-turbo",
         "stabilityai/stable-diffusion-xl-base-1.0",
         "/models/SDXL/official_pytorch/fp16/stable_diffusion_fp16/checkpoint_pipe",
+        "/models/SDXL/official_pytorch/fp16/stable_diffusion_fp16//checkpoint_pipe",
```
IanNod, line 198: We should try not matching hardcoded path's. Can we add a flag for the user to set which sd version to use for the model map so that downloaded model cards can be used with any path name?

models/turbine_models/custom_models/sd_inference/sd_pipeline.py
```diff
@@ -704,10 +843,35 @@ def numpy_to_pil_image(images):
...
     ireec_flags = {
-        "clip": args.ireec_flags + args.clip_flags,
+        "text_encoder": args.ireec_flags + args.clip_flags,
         "scheduler": args.ireec_flags,
         "unet": args.ireec_flags + args.unet_flags,
-        "vae_decode": args.ireec_flags + args.vae_flags,
+        "mmdit": args.ireec_flags + args.mmdit_flags,
+        "vae": args.ireec_flags + args.vae_flags,
+    }
+    devices = {
```
IanNod, line 852: Looks like we have to specify every submodels device? Necessary for compiling/running on multiple devices but can we have an overall optional flag to override and set all of these once if we are only running on one device?

models/turbine_models/custom_models/sd_inference/vae.py
```diff
@@ -161,20 +163,32 @@ def export_vae_model(
...
-        )
-    ]
+    if "stable-diffusion-3" in hf_model_name:
+        input_image_shape = (height, width, 3)
+        input_latents_shape = (batch_size, 16, height // 8, width // 8)
+    else:
+        input_image_shape = (batch_size, 3, height, width)
+        input_latents_shape = (batch_size, num_channels, height // 8, width // 8)
+
+    # encode_args = [
```
IanNod, line 186: commented code

models/turbine_models/tests/sdxl_test.py
```diff
@@ -10,16 +10,12 @@
 import shutil
```
IanNod, line 10: There has been a lot of pipelining and helper function's added. We should start working to integrate smaller unit tests for each case

models/turbine_models/custom_models/sd3_inference/diffusers_ref.py
monorimet: Doing a faithful comparison with diffusers reference is a larger problem -- we are really best off investing in getting real CLIP/FID scores with a validation dataset. This diffusers reference is really just a hold-over/sanity check for now; I don't even trust it to give us a decent baseline from CPU. We can leave this out for now but I'd rather keep it just to have something ready for comparison with diffusers on ROCM/CUDA

models/turbine_models/custom_models/sd3_inference/sd3_mmdit.py
monorimet: No, this timestep input gets expanded in the model to the batch dimension.

models/turbine_models/custom_models/sd_inference/sd_pipeline.py
monorimet: the pipeline does handle a string input for device and iree target. It's not really the best way to manage it at this point -- I'd rather the API consumer explicitly define targets for each submodel, but both modes work both in the CLI parser and the pipeline API.

models/turbine_models/tests/sdxl_test.py
monorimet: I see what you mean with this. Testing these gets pretty ugly without some context manager but these tests are definitely getting overloaded and need to be reduced to unit tests.

PR #845, opened 2024-09-18
(WIP) Add torchbench exports and benchmarks.

monorimet:
A README.md is included in the new directory for instructions on exporting the models. Not all are functional and the process requires a clone of the pytorch source code, currently, as well as a Linux distro for pytorch+rocm support (the torchbench modeling code relies on CUDA apis)

Once the exports are covered we can set up a simple benchmarking suite that runs the compiled .vmfbs and records benchmark results.

### DIFF
--- a/models/turbine_models/custom_models/torchbench/README.md
+++ b/models/turbine_models/custom_models/torchbench/README.md
@@ -1,0 +1,36 @@
+# SHARK torchbench exports and benchmarks
+
+### Setup
+
+ - pip install torch+rocm packages:
+```shell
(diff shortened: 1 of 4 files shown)


monorimet:
Notes on current state:

 - Need to separate out runtime functionality from export script
 - Start targeting torchbench integration with improvements to implementation, classing, statistics reporting, etc..
 - Integrate with SHARK-TestSuite CI
 - Work out upload/fetch of MLIR and weights so that this process can be used with containers running pytorch 2.3.0 without workarounds

PR #885, opened 2024-11-19
SDXL E2E Pipeline CI

saienduri:


.github/workflows/test_sdxl.yml
```diff
@@ -21,36 +23,70 @@ jobs:
...
+          pip install --no-compile --pre -e ${{ github.workspace }}/iree-turbine[testing]
+          pip install --no-compile --pre --upgrade -e models -r models/requirements.txt
+          pip uninstall torch torchvision torchaudio -y
+          pip install https://download.pytorch.org/whl/nightly/pytorch_triton_rocm-3.0.0%2B21eae954ef-cp311-cp311-linux_x86_64.whl
+          pip install https://download.pytorch.org/whl/nightly/rocm6.1/torch-2.5.0.dev20240710%2Brocm6.1-cp311-cp311-linux_x86_64.whl
+          pip install https://download.pytorch.org/whl/nightly/rocm6.1/torchvision-0.20.0.dev20240711%2Brocm6.1-cp311-cp311-linux_x86_64.whl
+          pip install https://download.pytorch.org/whl/nightly/rocm6.1/torchaudio-2.4.0.dev20240711%2Brocm6.1-cp311-cp311-linux_x86_64.whl
+          pip uninstall iree-compiler iree-runtime iree-base-compiler iree-base-runtime -y
+      
+      - name: Build IREE
```
ScottTodd, line 64: FYI you can fetch prebuilt Python wheels from GitHub artifacts where pkgci.yml has run, if you want builds newer than nightly releases. Should save you some time and disk space.

(downstream should never have to build the compiler from source unless extending with some extra source code that needs to be included in the build)

.github/workflows/test_sdxl.yml
```diff
@@ -21,36 +23,70 @@ jobs:
...
+          source iree_venv/bin/activate
+          pip install --upgrade pip
+          pip install -r runtime/bindings/python/iree/runtime/build_requirements.txt
+          cd $GITHUB_WORKSPACE/iree
+          git submodule update --init
+          cmake -S . -B build-release \
+            -G Ninja -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_C_COMPILER=`which clang` -DCMAKE_CXX_COMPILER=`which clang++` \
+            -DIREE_HAL_DRIVER_CUDA=OFF \
+            -DIREE_EXTERNAL_HAL_DRIVERS="rocm" \
```
nithinsubbiah, line 76: `-DIREE_HAL_DRIVER_HIP=ON -DIREE_HAL_DRIVER_HIP=ON` for compiler and runtime backend

.github/workflows/test_sdxl.yml
```diff
@@ -21,36 +23,70 @@ jobs:
...
-          # from non default locations first. Installing the PyTorch CPU
-          # wheels saves multiple minutes and a lot of bandwidth on runner setup.
-          pip install --no-compile --index-url https://download.pytorch.org/whl/cpu \
-            -r ${{ github.workspace }}/iree-turbine//pytorch-cpu-requirements.txt
-          pip install --no-compile --upgrade -r ${{ github.workspace }}/iree-turbine/requirements.txt
-          pip install --no-compile -e ${{ github.workspace }}/iree-turbine/[testing,torch-cpu-nightly]
-          pip install --no-compile --upgrade -r models/requirements.txt
-          pip install --no-compile -e models
+          pip install --no-compile -r ${{ github.workspace }}/iree-turbine/pytorch-cpu-requirements.txt
+          pip install --pre --upgrade -r ${{ github.workspace }}/iree-turbine/requirements.txt
```
nithinsubbiah, line 54: Do we need to install `iree-turbine` explicitly since it's listed as part of `models/requirements.txt`?

nithinsubbiah (changes requested):

.github/workflows/test_sdxl.yml
saienduri: The goal of this CI is to test hourly on TOM of everything, so it's easy to pinpoint where something broke. If there's a way to pull the latest PkgCI wheels at any point that would be great. But, it would have to be from TOM not on a PR.

ScottTodd: Yeah, https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28 query workflow runs triggered by `push` events on the `main` branch, then get the artifacts: https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#get-an-artifact. I don't think we have the one-liner (or... more like 10-50 liner...) to do that fetching, but it would be good to have it.

.github/workflows/test_sdxl.yml
saienduri: IREE build no longer needed

.github/workflows/test_sdxl.yml
saienduri: Yeah not needed

.github/workflows/test_sdxl.yml
saienduri: Oooh that's perfect! I wrote a script for this, and it's working (https://github.com/nod-ai/SHARK-ModelDev/pull/885/files#diff-0c1eaf69a654192f737bb2da8e399fb90d8cbade47b0b98d1ea2056617b5ff5a). Thanks

nithinsubbiah (approved):
Thanks for putting this up

PR #901, opened 2025-01-21
Pin `action/checkout` action

marbre:
Pins as suggested by OpenSSF Scorecard, see
https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies.


ScottTodd (approved):

marbre:
CI failures are unrelated.

PR #902, opened 2025-01-21
Update GH actions with Dependabot

marbre:
All actions are pinned with hases as suggested by OpenSSF Scorecard, as soon as #901 lands. Actions can be upgraded with Dependabot, see https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+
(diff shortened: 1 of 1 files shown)


ScottTodd:
You might need to adjust repository settings to enable this after merge.

image

